PR #3, opened 2026-04-07
feat: add permission relay support

SamSchimek:
## Summary

Adds permission relay to the Slack channel plugin, allowing users to approve or deny Claude Code tool actions remotely from their phone via Slack.

- Declares `claude/channel/permission` capability
- Handles `permission_request` notifications — posts formatted prompt to the last active channel/thread
- Intercepts `y/yes/n/no + 5-char code` replies as verdicts before normal chat delivery
- Acks processed verdicts with ✅ reaction
- Tracks last active channel/thread for prompt routing

Uses the permission relay protocol documented at [code.claude.com/docs/en/channels-reference](https://code.claude.com/docs/en/channels-reference#relay-permission-prompts).

## Test plan

- [x] `bun test` — 52 tests pass, no regressions
- [x] `bun build --no-bundle server.ts` — compiles clean
- [x] Tested end-to-end: permission prompt appears in Slack, `y <code>` reply approves the action, ✅ reaction confirms

Co-Authored-By: Sam Schimek <sam@backyard.chat>
Co-Authored-By: Claude Code <noreply@anthropic.com>

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -14,6 +14,7 @@ import {
   CallToolRequestSchema,
   ListToolsRequestSchema,
 } from '@modelcontextprotocol/sdk/types.js'
+import { z } from 'zod'
 import { SocketModeClient } from '@slack/socket-mode'
 import { WebClient } from '@slack/web-api'
 import { homedir } from 'os'
@@ -176,6 +177,10 @@ function assertSendable(filePath: string): void {
 // Track channels that passed inbound gate (session-lifetime cache)
 const deliveredChannels = new Set<string>()
 
+// Track last active channel/thread for permission relay
+let lastActiveChannel = ''
+let lastActiveThread: string | undefined
+
 function assertOutboundAllowed(chatId: string): void {
   libAssertOutboundAllowed(chatId, getAccess(), deliveredChannels)
 }
@@ -223,7 +228,10 @@ const mcp = new Server(
   { name: 'slack', version: '0.1.0' },
   {
     capabilities: {
-      experimental: { 'claude/channel': {} },
+      experimental: {
+        'claude/channel': {},
+        'claude/channel/permission': {},
+      },
       tools: {},
     },
     instructions: [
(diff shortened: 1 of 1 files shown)


Lingnik:
FYSA @SamSchimek - #4

jeremylongshore:
Hey @SamSchimek — thanks for this, the permission relay is exactly the feature this plugin needed and you nailed the core flow. I want to merge this as the foundation, then @Lingnik's Block Kit work in #4 can land on top.

Three things to address before I can merge — all fairly quick:

**1. Add `assertOutboundAllowed()` before posting the permission prompt** (~L565)
The relay posts to `targetChannel` without the outbound gate check. Every other outbound Slack call in the codebase goes through this guard — the permission prompt should too. One-liner:
```ts
assertOutboundAllowed(targetChannel)
```

**2. Text reply should check the sender is the session owner** (~L617)
Right now any user whose message passes the channel gate can send `y abcde` and approve a tool call. The intent is that only the session owner (in `access.allowFrom`) can approve. Add a check before forwarding the verdict:
```ts
if (!access.allowFrom.includes(ev['user'] as string)) return
```

**3. Escape `tool_name` and `description` before posting** (~L569)
These come from Claude Code so they're likely safe, but defense-in-depth matters here — a `tool_name` containing `<!channel>` would ping everyone. Escape `<`, `>`, `&` before interpolation.

No new tests required for this PR — the Block Kit PR (#4) is the right place to add coverage for the full permission path.

`zod`: it's a transitive dep of `@modelcontextprotocol/sdk` already, so the import works, but please add it to `package.json` dependencies so it's explicitly declared.

These are all small changes. Once they're in I'll merge promptly. Really appreciate you kicking this off.

jeremylongshore:
Merged! I added a security hardening commit on top of your work before merging:

- **Outbound gate** — added `assertOutboundAllowed()` before posting permission prompts, so the relay respects the same security boundary as every other outbound Slack call
- **Owner-only approval** — text-based replies (`y abcde`) now check that the sender is in `access.allowFrom`, so only the session owner can approve tool calls (not any channel member)
- **mrkdwn escaping** — `tool_name` and `description` are escaped before interpolation to prevent Slack mention injection (`<!channel>` etc.)
- **TS2589 fix** — zod v3/v4 type recursion workaround so typecheck passes clean
- **zod declared** — added to `package.json` as explicit dependency

Your commit is in main with full attribution. @Lingnik's Block Kit PR (#4) will build on top of this. Thanks for kicking off the permission relay — it's a great addition.

PR #4, opened 2026-04-07
feat: add permission relay support (by @SamSchimek) with block kit support for approvals

Lingnik:
# Summary

Extends the permission relay from #3 with Slack Block Kit interactive buttons, so users can tap Allow/Deny on their phone instead of typing y abcde.

Built on top of @SamSchimek's #3 — includes his permission relay commit as the base.

# Block Kit buttons

 - Allow (green), Deny (red), Details buttons rendered via Block Kit
 - Details expands input_preview as plain_text to prevent mrkdwn injection
 - Messages update in-place to show outcome and remove buttons
 - Non-owner button clicks get ephemeral rejection: "Only the session owner can approve or deny tool calls"

# Security hardening (addresses qodo review on #3)

 - Outbound gate enforced on permission relay messages (#3 review item 1)
 - zod added as explicit dependency (#3 review item 2)
 - request_id format validated before use in action_id fields
 - tool_name / description escaped to prevent mrkdwn injection
 - Duplicate verdict guard on both button and text paths

# Other fixes

 - Fix TS2589 type recursion between zod and MCP SDK generics (typecheck now passes)
 - Fix "denyed" → "denied" typo
 - Add try/catch on interactive handler (matches message/app_mention pattern)
 - TTL-based cleanup for pendingPermissions map (5-min expiry)

# Slack App requirement

Interactivity must be enabled in app settings for buttons to work. Socket Mode delivers interaction payloads — no public URL needed.

# Known limitation

When a permission prompt is resolved from the terminal (not via Slack), the Block Kit buttons remain in Slack. Clicking them shows "Already resolved." A permission_resolved notification in the channels protocol would fix this, but that's an upstream change -- in Claude Code.

# Test plan

- bun test — 52 tests pass, no regressions
- bun run typecheck — passes clean
- End-to-end: permission prompt renders buttons in Slack
- Allow button delivers verdict, message updates
- Deny button delivers verdict, message updates
- Details button expands input_preview as plain_text
- Non-owner click shows ephemeral rejection
- Text-based y/n <code> fallback works
- Terminal approval works in parallel with Slack buttons
- Duplicate click shows "Already resolved"

 Co-Authored-By: Sam Schimek sam@backyard.chat
 Co-Authored-By: Claude Opus 4.6 (1M context) noreply@anthropic.com

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "@anthropic/claude-channel-slack",
@@ -7,6 +8,7 @@
         "@modelcontextprotocol/sdk": "^1.0.0",
         "@slack/socket-mode": "^2.0.0",
         "@slack/web-api": "^7.0.0",
+        "zod": "^3.24.0",
       },
       "devDependencies": {
         "@types/bun": "^1.0.0",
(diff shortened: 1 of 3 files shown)


jeremylongshore:
Hey @Lingnik — good work building on Sam's relay with Block Kit buttons. The interactive UX is well thought out and the security hardening (mrkdwn escaping, request_id validation, duplicate verdict guard) shows you engaged with the architecture seriously.

Here's my plan: I'm going to merge #3 first with a few fixes so Sam gets merge credit for the foundation. Once that's in main, could you rebase this PR onto main? That'll drop the duplicate first commit and make your PR a clean "add Block Kit interactive layer" diff.

After rebase, a few things to address:

**1. Text-based approval must check session owner** (~L835)
Same issue as #3 — the button path restricts to `access.allowFrom`, but the text path doesn't. Any channel member could `y abcde` to approve a tool call.

**2. Move `pendingPermissions.delete()` after `mcp.notification()`** (~L759, ~L854)
If the notification fails, the entry is already gone and the user can't retry. Delete after successful send.

**3. Add test coverage**
The existing suite covers every security function. The new permission path needs at least: `escMrkdwn()` and `pruneStalePermissions()` tests (move both to `lib.ts` so they're testable in isolation — matches the project convention), and a test that the owner-check logic works.

**4. Call `pruneStalePermissions()` in the interactive + text paths too**
Currently it only runs on new `permission_request` arrivals, so expired entries remain actionable.

Happy to discuss any of this. Thanks to you and @SamSchimek both — this is a meaningful addition to the plugin.

jeremylongshore:
Merged! I rebased onto main (after merging #3), resolved conflicts, and applied security fixes during the merge:

**Fixes applied:**
- **Owner-only text approval** — text-based replies (`y abcde`) now check `access.allowFrom`, matching the button path's owner restriction. Without this, any channel member could approve tool calls.
- **Delete after send** — moved `pendingPermissions.delete()` to *after* `mcp.notification()` succeeds, in both button and text paths. Previously if the notification failed, the request was lost and couldn't be retried.
- **TTL enforcement at decision time** — added `pruneStalePermissions()` in the interactive handler and text-reply path, not just on new permission requests. Expired entries can no longer be acted on.
- **Outbound gate direct throw** — changed from try/catch (which silently swallows) to direct `assertOutboundAllowed()` call, matching the pattern used everywhere else in the codebase.

Your Block Kit implementation, mrkdwn escaping, request_id validation, and duplicate verdict guard all shipped as-is — good work. Both your and @SamSchimek's commits are in main with full attribution.

PR #5, opened 2026-04-11
Security hardening: six fixes from a pre-deployment review

maui-99:
## Summary

Six security fixes surfaced by a blind pre-deployment security review against v0.2.0. Each commit is scoped to a single issue and ships with tests (or docs that match the new guards). No user-facing defaults change in this PR — each fix is a strict tightening over the current behavior, so the upgrade path is a no-op for existing installs.

Happy to split this further, rework individual patches, or drop any commit the maintainer doesn't want — filing as one PR for reviewability.

## Findings addressed

1. **`assertSendable` was state-dir-scoped only (CRITICAL).** The previous guard in `lib.ts` only blocked paths outside `CCSC_STATE_DIR`, which still allowed `read_file`/`edit_file` callers to touch anything else on disk. Replaced with an explicit allowlist (user-configured `CCSC_READ_ALLOW` / `CCSC_EDIT_ALLOW` roots) plus a denylist for common credential paths (`~/.ssh`, `~/.aws`, `~/.config/gcloud`, `.env`, etc.). Access decisions are logged. Tests cover traversal, symlinks, and denylist bypass attempts.

2. **Floating dep ranges + no frozen lockfile (CRITICAL).** `package.json` used `^` ranges and `bun install` in `start` did not pass `--frozen-lockfile`, so a compromised transitive could land on next boot. Pinned all runtime and dev deps to exact versions and added `--frozen-lockfile` to the `start` script so resolution drift fails boot instead of silently updating.

3. **`read_file` / `edit_file` / `react` / `download_attachment` bypassed the allowlist (HIGH).** The new `assertSendable` guard was not wired into the tool handlers in `server.ts`. Gated all four tools through the outbound allowlist so prompt-injected tool calls from Slack content cannot escape the configured roots.

4. **MCP instructions didn't mention the allowlist (docs follow-up).** Updated the MCP server instructions string so Claude is told up-front which paths it can read/edit, matching the runtime guard. Prevents the model from attempting out-of-bounds reads and getting confusing tool errors.

5. **Slack display names injected raw into the model context (MEDIUM).** `user.profile.display_name` flowed into the prompt context unescaped, giving any workspace member a prompt-injection surface by renaming themselves. Added a sanitizer that strips control chars, zero-width, bidi overrides, and markdown/tool-call lookalikes, with unit tests for each class.

6. **`access.json` tmp file had a mode race (LOW).** The atomic-write path created the tmp file at default umask and then `chmod`ed it to `0600`, leaving a brief window where the file was world-readable. Fixed by passing `mode: 0o600` to `writeFileSync` so the file is created with the restricted mode atomically.

## Test plan

- [x] `bun install --frozen-lockfile` succeeds
- [x] `bun test` — 86 pass / 0 fail on this branch
- [x] Each commit individually reviewed for scope
- [x] End-to-end smoke test on a real Slack workspace (bot boots, responds in allowed channel, refuses out-of-allowlist file read, refuses out-of-allowlist file edit, pairing flow still works when re-enabled)

Happy to discuss any of the individual patches if you'd like them reworked.


jeremylongshore:
## Merged — thank you, Ian

This is outstanding work. Every finding maps to a real attack vector, the code is well-structured (pure functions in `lib.ts`, wiring in `server.ts`), and the 34 new tests are thorough — including real-filesystem symlink fixtures with proper cleanup.

### What I changed during merge

**Restored `start:node` script.** The original PR removed `"start:node": "npx tsx server.ts"` from `package.json`. This is a documented alternative runtime for Node.js users (referenced in README), so I added it back in a separate commit on the branch before merging. Everything else shipped exactly as submitted.

### Notes for the changelog

- **BREAKING:** `assertSendable` policy changed from permissive (allow everything except state dir) to restrictive (deny everything except inbox + configured `SLACK_SENDABLE_ROOTS`). Users who have Claude attach arbitrary files via `reply` will need to set `SLACK_SENDABLE_ROOTS` in their `.env`.
- Dependency versions pinned to exact (removed `^` ranges). Future updates require explicit bumps.
- `zod` pinned to `3.25.76` — the MCP SDK ships with `zod@4.x` internally, but this pin is correct for our direct usage.

### PR description note

The PR body mentions "read_file / edit_file bypassed the allowlist" — these MCP tools don't exist in this server. The actual fixes (which are correct in the code) are for `react`, `edit_message`, `fetch_messages`, and `download_attachment`. Minor description issue, code is spot-on.

---

52 → 86 tests. Six real vulnerabilities closed. Exceptional contribution. 🎯

PR #7, opened 2026-04-15
fix: terminate MCP server cleanly on client disconnect

jinsung-kang:
## Summary
- MCP server leaked a zombie `bun server.ts` process every time a Claude Code session ended. The SDK's `StdioServerTransport` only listens for stdin `data`/`error`, not `end`/`close`, so the pipe EOF was silently ignored — and the Socket Mode WebSocket kept the event loop alive indefinitely.
- Added an explicit `shutdown()` path wired to four triggers: `process.stdin` `end`/`close` (parent hangup), `transport.onclose` (SDK-side signal), and `SIGINT`/`SIGTERM` (interactive/manual). `shutdown()` disconnects the Socket Mode client, closes the MCP server, and is guarded by a 3s `unref()`-ed forced-exit timer so the safety net itself doesn't keep the loop alive.
- Re-entry is blocked by a `shuttingDown` flag so overlapping triggers (e.g. stdin EOF firing right after SIGTERM) don't double-teardown.

## Test plan
- [x] `bun run typecheck` passes
- [x] `bun test` — 86 pass, 0 fail
- [ ] Manually: start a Claude Code session that loads this MCP server, end the session, confirm `pgrep -fl "bun server.ts"` returns empty
- [ ] Manually: run `bun server.ts` standalone, send SIGINT (Ctrl+C), confirm clean exit within 3s
- [ ] Manually: run `bun server.ts` standalone, send SIGTERM (`kill <pid>`), confirm clean exit within 3s

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -985,6 +985,47 @@ socket.on('app_mention', async ({ event, ack }) => {
 })
 
 // ---------------------------------------------------------------------------
+// Shutdown
+// ---------------------------------------------------------------------------
+//
+// Without this, the process turns into a zombie when Claude Code disconnects.
+// The MCP SDK's StdioServerTransport only listens for stdin `data`/`error` —
+// not `end`/`close` — so EOF on the pipe is silently ignored. Meanwhile the
+// Socket Mode WebSocket keeps pinger/reconnect timers alive, holding the
+// event loop open indefinitely. We hook stdin directly, plus SIGINT/SIGTERM,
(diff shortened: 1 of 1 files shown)


jeremylongshore:
Merging — thanks for the clean diagnosis and the careful edge-case handling (re-entry guard, unref'd timer, unconnected-socket case all covered). If you ever want to come back and split `shutdown()` into a testable controller in `lib.ts`, I'd happily land that too.

PR #8, opened 2026-04-15
fix: dedupe duplicate event delivery from message + app_mention

CaseyMargell:
## Summary

The server subscribes to both `message` and `app_mention` events (`server.ts:966,977`), routing both through `handleMessage`. In Slack's Events API, `app_mention` isn't a replacement for `message` — it's an *additional* event that fires alongside a regular `message` event when the bot is @-mentioned. Result: for any @-mentioned channel message, `handleMessage` runs twice, causing doubled inference, doubled replies, and doubled token cost.

Slack's own event redelivery (on slow acks) causes the same symptom.

## Repro

1. Add the bot to a channel and opt that channel in under `access.json` with `requireMention: false`
2. Post a message from Slack that @-mentions the bot
3. Observe: `handleMessage` processes the event twice; two replies get posted

The comment at `server.ts:976` ("used in channels with requireMention") suggests the `app_mention` subscription was thought necessary when `requireMention` is true. It isn't — the regular `message` event carries the mention text and `isMentioned()` reads that directly, so `app_mention` is redundant for bots that are members of the channel.

## Approach

Adds `isDuplicateEvent()` to `lib.ts` as a pure function, matching the lib/server split described in `CLAUDE.md`. Keyed on `(channel, ts)` — Slack's unique identifier for a message. The cache is caller-supplied (a `Map`), so the function stays pure and is directly testable. Prunes expired entries on every check; TTL is 60s.

Wired into `handleMessage` at the top so both subscriptions share one process-lifetime cache.

Considered alternative: delete the `app_mention` handler outright. Not taken because dedup also defends against Slack's own event redelivery, which happens occasionally even with only one subscription. Happy to switch to the delete-the-handler approach if you prefer fewer moving parts — it would be a smaller diff.

## Test plan

- [x] `bun test` — 95 pass, 0 fail (86 existing + 9 new)
- [x] New tests cover: first-seen, repeat-within-TTL, repeat-after-TTL, cross-channel independence, same-channel different-ts independence, undedupable events (missing channel or ts), TTL-based pruning, and the exact message+app_mention dual-delivery scenario this PR addresses

## Notes

Happy to split into "lib function" + "server wiring" commits, swap to the alternative approach above, or drop any part of this if the shape doesn't fit. Filed as one commit for reviewability.

### DIFF
--- a/lib.ts
+++ b/lib.ts
@@ -454,3 +454,51 @@ function isMentioned(event: Record<string, unknown>, botUserId: string): boolean
   const text = (event['text'] as string | undefined) || ''
   return text.includes(`<@${botUserId}>`)
 }
+
+// ---------------------------------------------------------------------------
+// Event deduplication
+// ---------------------------------------------------------------------------
+
+/** How long a seen (channel, ts) pair stays in the dedup cache. */
+export const EVENT_DEDUP_TTL_MS = 60_000
+
+/**
+ * Detect and record a duplicate Slack event.
+ *
+ * Slack subscribes bots to both `message` and `app_mention` events. A channel
+ * message that @-mentions the bot arrives via BOTH subscriptions, so the
+ * server's handler runs twice for the same Slack message. This also protects
+ * against Slack's own event redelivery when an ack is slow or missed.
+ *
+ * Dedup key is (channel, ts) — Slack's own unique identifier for a message.
+ *
+ * Mutates `seen` in place: prunes expired entries on every call, then
+ * records this event with expiry at `now + ttlMs`. Returns true when the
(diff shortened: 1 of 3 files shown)


jeremylongshore:
Merging — thanks for the detailed PR body and the 9 new tests, both helped a lot. FYI for future reference: Slack's docs say `app_mention` is redundant when the bot is a channel member (which this repo requires via the gate), so a future simplification could just drop that subscription per your own alt proposal. Keeping the dedup is fine either way — Slack's retry-on-slow-ack behavior is real, so this provides defense-in-depth. Tracking the scaling follow-ups in a separate issue.

PR #33, opened 2026-04-18
feat(gate): per-channel allowBotIds for opt-in cross-bot delivery

### LINKED ISSUE #27: Proposal: per-channel opt-in for cross-bot message delivery
## Use case

Multi-agent coordination in shared channels. When two Claude Code instances (e.g., an ops-monitor bot and an engineering bot) need to coordinate incident response in a shared `#incidents` channel, the current blanket `bot_id` drop in `gate()` (lib.ts:385) prevents them from seeing each other's messages. An @mention from one bot to the other is silently dropped before any access-control check runs.

This came up while building an autonomous incident response flow: ops-monitor detects an issue, posts to `#incidents`, and @-mentions the engineering bot. The engineering bot never sees the message.

## Current behavior

```ts
// lib.ts gate() — line 385
if (ev['bot_id']) return { action: 'drop' }
```

Every message with a `bot_id` field is dropped unconditionally. This is a good default — it prevents bot-loop noise and unexpected interactions with third-party integrations. But it also blocks intentional multi-agent coordination.

## Proposal

Add an optional `allowBotMessages` flag to the per-channel policy in `access.json`:

```json
{
  "channels": {
    "C_INCIDENTS": {
      "requireMention": false,
      "allowFrom": ["U_OPS_BOT", "U_ENG_BOT", "U_HUMAN"],
      "allowBotMessages": true
    }
  }
}
```

When `allowBotMessages` is `true` for a channel:
- Bot messages pass through the gate (instead of being dropped at step 1)
- Self-echo is still filtered (`bot_id` present AND `user === botUserId` → drop)
- `allowFrom` still scopes which user/bot IDs are accepted
- `requireMention` still applies

When `false` or absent (default): current behavior — all bot messages dropped. Zero impact on existing installs.

### Gate change

```ts
// Replace blanket drop with self-echo filter + policy check
if (ev['bot_id']) {
  // Always drop our own echoes
  if (ev['user'] === opts.botUserId) return { action: 'drop' }
  // For channel messages, check if the channel opts in to bot messages
  const channel = ev['channel'] as string
  const policy = access.channels[channel]
  if (!policy?.allowBotMessages) return { action: 'drop' }
  // Fall through to normal access-control checks (allowFrom, requireMention)
}
```

### Safety properties

- **Default-safe.** No `allowBotMessages` = current behavior. Existing installs are unaffected.
- **Self-echo always filtered.** Prevents reply loops regardless of config.
- **Scoped by `allowFrom`.** Even with `allowBotMessages: true`, only the listed bot user IDs can trigger. Random third-party integrations added to the channel are still dropped.
- **Dedup protects against double-delivery.** The event dedup from #8 ensures that even if both `message` and `app_mention` fire for a bot's @-mention, `handleMessage` processes it once.
- **DMs unaffected.** Bot DMs are still dropped — DM channel IDs are never in the `access.channels` map, so `policy?.allowBotMessages` evaluates to falsy before the DM handling path is reached.

## Context

- Issue #8 landed the event dedup that makes this safer (no double-delivery of bot messages)
- Issue #11 already tracks "bot-to-bot chatter" as a scaling scenario for the dedup cache
- I have a working implementation on my fork ([branch](https://github.com/CaseyMargell/claude-code-slack-channel/tree/casey/allow-cross-bot-messages)) with a simpler unconditional version. Happy to rework it to the per-channel config approach above and submit a PR if this direction looks right.

## Access.md update

Would add documentation for the new `allowBotMessages` field alongside `requireMention` and `allowFrom` in the channel policy schema.

CaseyMargell:
## Summary

Implements the five design refinements from #27 for per-channel cross-bot message delivery. Bot messages are dropped by default (no behavior change for existing installs). Channels opt in by listing specific bot user IDs in a new `allowBotIds` field on the channel policy.

## Design refinements (all five from #27)

**1. `allowBotIds: string[]` replaces proposed `boolean`.** Explicit bot user ID whitelist on `ChannelPolicy`. Absent or empty = all bots dropped. Auditable, fails closed, survives an operator forgetting to lock down `allowFrom`.

**2. Self-echo triple-check.** Captures `selfBotId` and `selfAppId` from `auth.test` at startup alongside `botUserId`. Drops if ANY of `{bot_id === selfBotId, bot_profile.app_id === selfAppId, user === botUserId}` matches. Covers `as_user=false` and multi-workspace payload variants.

**3. Permission reply guard in `gate()`.** Runs `PERMISSION_REPLY_RE` against peer-bot text before delivery and drops matches. Belt-and-suspenders with the existing global `allowFrom` checks at `server.ts:704` (button handler) and `:876` (text permission reply), which already block peer bots from approving tool calls. This gate-level check prevents regression if that guard is ever loosened.

**4. Audit log.** `console.error('[slack] bot message delivered', { bot_id, user, channel, ts })` on every delivered bot message for multi-agent flow diagnostics.

**5. System prompt hardening.** One sentence in the MCP instructions block noting peer-bot messages carry the same prompt-injection risk as human messages.

## Permission relay blast radius

The existing global-`allowFrom` guards at `server.ts:704` (button click handler) and `server.ts:876` (text permission reply) gate on `access.allowFrom`, not the channel policy. Peer bots listed in a channel's `allowBotIds` but not in the global `allowFrom` cannot approve permission prompts via either path. The `PERMISSION_REPLY_RE` guard in `gate()` adds a third layer that catches permission-shaped text before it even reaches the delivery branch.

## Moved constant

`PERMISSION_REPLY_RE` moved from `server.ts` to `lib.ts` (pure constant, no behavior change) so `gate()` can reference it for refinement #3. `server.ts` imports it from `lib.ts`.

## Test plan

- [x] `bun test` — 104 pass, 0 fail (95 existing + 9 new)

New `allowBotIds` test block in `server.test.ts` (items **4, 5, 7, 9 are the non-negotiable security-critical edges**):

1. Drops bot message when channel has no `allowBotIds` (default-safe regression guard)
2. Drops bot message when bot user_id not in `allowBotIds` (whitelist enforcement)
3. Delivers bot message when bot user_id in `allowBotIds` AND channel `allowFrom` includes it
4. **Drops self-echo via `bot_id` match even when `allowBotIds` includes our `botUserId`**
5. **Drops self-echo when `ev.user` is missing but `bot_profile.app_id` matches**
6. Drops bot message in DM channel even with `allowBotIds` set on a different channel
7. **Drops peer-bot message matching `PERMISSION_REPLY_RE` (e.g., `y abcde`)**
8. `requireMention` still applies to peer-bot messages
9. **Peer bot not in global `access.allowFrom` cannot trigger permission relay** — verifies gate-level `PERMISSION_REPLY_RE` check catches it before the delivery branch

Closes #27

---

Casey made me do it
\- claude

### DIFF
--- a/lib.ts
+++ b/lib.ts
@@ -19,6 +19,11 @@ export const MAX_PENDING = 3
 export const MAX_PAIRING_REPLIES = 2
 export const PAIRING_EXPIRY_MS = 60 * 60 * 1000 // 1 hour
 
+/** Matches permission relay replies (e.g. "y abcde", "no xyzwq").
+ *  Used in gate() to block peer-bot messages that look like permission
+ *  approvals, and in server.ts to route human permission replies. */
+export const PERMISSION_REPLY_RE = /^\s*(y|yes|n|no)\s+([a-km-z]{5})\s*$/i
+
 // ---------------------------------------------------------------------------
 // Types
 // ---------------------------------------------------------------------------
@@ -28,6 +33,9 @@ export type DmPolicy = 'pairing' | 'allowlist' | 'disabled'
 export interface ChannelPolicy {
   requireMention: boolean
   allowFrom: string[]
+  /** Opt-in list of bot user IDs allowed to deliver messages in this channel.
+   *  Absent or empty = all bot messages dropped (default-safe). */
+  allowBotIds?: string[]
 }
 
 export interface PendingEntry {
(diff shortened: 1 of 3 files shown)


jeremylongshore (approved):
Approved. CODEOWNERS human review — `lib.ts` + `server.ts`.

## Audit

- **Self-echo triple-check** — correct. `bot_id === selfBotId || bot_profile.app_id === selfAppId || user === botUserId`. Fails closed on missing-user payload variants.
- **`allowBotIds` opt-in** — requires `botUser` present AND in the list. `!policy?.allowBotIds?.length || !botUser || !includes(botUser)` is the right short-circuit.
- **`PERMISSION_REPLY_RE` guard in `gate()`** — executes before fall-through, belt-and-suspenders with server.ts:704 and :876 as requested.
- **Fall-through** preserves existing `requireMention` and `allowFrom` semantics for bot messages. Test 8 confirms `requireMention`.
- **DMs unaffected** — test 6 validates. DM channel IDs aren't in `access.channels`, so `policy?.allowBotIds?.length` short-circuits to drop.
- **Audit log** at `server.ts:872-878` fires inside `case 'deliver'`, only on accept. Matches spec.
- **System-prompt addendum** at `server.ts:277` — one sentence on peer-bot prompt-injection risk. Correctly positioned inside the MCP instructions block.
- **Constant move** (`PERMISSION_REPLY_RE` from `server.ts` to `lib.ts`) — clean refactor, pure constant, no behavior change. Shared between `gate()` and permission-reply routing.
- **Startup identity log** at `server.ts:1063` — useful for operator diagnostics without leaking secrets.

## Tests

All 9 cases match the spec. Items 4, 5, 7, 9 (the non-negotiable security-critical ones) assert the right invariants. Test 7 also verifies the regex directly — good belt-and-suspenders.

104 pass / 0 fail locally.

## Nit (non-blocking)

`access.channels[channel]` is looked up twice in the bot_id block and again in the channel-handling block. Could hoist once at the top of `gate()` later; not worth rebuilding for.

## CI

- Typecheck ✅
- CodeQL ✅
- javascript-typescript analysis ✅
- gemini-review ❌ — **workflow auth config issue, not a code finding.** Separate fix incoming, does not block this PR.

Ready to merge.

Jeremy made me do it
-claude

PR #63, opened 2026-04-19
feat: add Node.js/tsx as alternative runtime

gog5-ops:
## Summary
- Switch `.mcp.json` from Bun to tsx (Node.js) as MCP server command
- Add `tsx` as runtime dependency in `package.json`
- Fixes Bun event loop bug where Socket Mode `message` events stop firing when `StdioServerTransport` is also active on stdin

## Context
When running under Bun, the Socket Mode WebSocket `message` handler never fires once `StdioServerTransport` attaches its own `data` listener to `process.stdin`. The same code works correctly under Node.js/tsx. Isolated with minimal test scripts confirming the issue is Bun-specific.

## Test plan
- [ ] Verify `bun test` still passes (tests use `bun:test`, unaffected)
- [ ] Verify MCP server starts and connects to Slack via Socket Mode under tsx
- [ ] Verify DM delivery and reply round-trip works
- [ ] Verify pairing flow works end-to-end

### DIFF
--- a/.mcp.json
+++ b/.mcp.json
@@ -1,8 +1,8 @@
 {
   "mcpServers": {
     "slack": {
-      "command": "bun",
-      "args": ["run", "--cwd", "${CLAUDE_PLUGIN_ROOT}", "--shell=bun", "--silent", "start"]
+      "command": "${CLAUDE_PLUGIN_ROOT}/node_modules/.bin/tsx",
+      "args": ["${CLAUDE_PLUGIN_ROOT}/server.ts"]
     }
   }
 }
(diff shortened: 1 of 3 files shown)


jeremylongshore:
Thanks for this @gog5-ops — the Bun event-loop claim is specific and the minimal fix matches CLAUDE.md's "Node.js/Docker as alternatives" stance, so happy to take it.

CI was only failing because `bun.lock` wasn't regenerated alongside the new `tsx` dependency (CI runs `bun install --frozen-lockfile`). I tried to push a lockfile regeneration straight to your fork via maintainer-edits, but the push was rejected on permission. So I've opened #65 with your original commit preserved as-is (attributed to you) plus a second commit that regenerates the lockfile and rebases on main.

Closing this PR in favor of #65. Credit stays on your commit. Thanks again.

Jeremy made me do it
-claude

jeremylongshore:
Superseded by #65 (same change, lockfile regenerated, rebased on main). Thanks @gog5-ops 🙏

PR #66, opened 2026-04-19
feat(supervisor): implement quiesce(key) (ccsc-xa3.3, Epic 32-B)

jeremylongshore:
## Summary

Second runtime under Epic 32-B. Replaces the staged stub on \`SessionSupervisor.quiesce()\` with a real graceful-drain implementation and grows \`ConcreteHandle\` with the package-private \`beginQuiesce\` / \`beginWork\` / \`endWork\` surface that the tool-call path (\`ccsc-xa3.15\`) will plug into.

### Behaviour

- **State transition** — \`supervisor.quiesce(key)\` looks up the live handle and calls \`handle.beginQuiesce()\`, flipping state \`active\` → \`quiescing\` synchronously. Missing key → silent no-op (no log) since there is no session to drain.
- **Drain** — returns a promise that resolves when the in-flight map reaches zero entries. If the map is already empty, resolution is queued on the next microtask (not inline) so callers cannot accidentally depend on synchronous resolution.
- **Idempotent (concurrent)** — parallel \`quiesce()\` calls share one drain promise via the handle. Each call still emits one \`session.quiesce\` log line because every quiesce *request* is audit-worthy.
- **Idempotent (post-drain)** — quiesce on an already-drained handle returns the cached resolved promise without re-entering the FSM edge.
- **Log** — \`session.quiesce\` with \`{ channel, thread, inflight }\`.
- **Graceful, not forceful** — \`beginQuiesce()\` does NOT abort in-flight work. That's the shutdown path (\`ccsc-xa3.14\`), which will call \`.abort()\` on every entry.

### ConcreteHandle growth

- \`beginWork(requestId)\` allocates and registers an AbortController. Throws on duplicate \`requestId\` to catch bookkeeping bugs before they mask double-resolve drains.
- \`endWork(requestId)\` removes the controller and resolves the drain if applicable. Unknown \`requestId\` is a deliberate no-op so crash-recovery double-end is not fatal.

### Scope notes

- **Not in this bead:** Quiescing → Deactivating (\`ccsc-xa3.14\`), Quiescing → Active cancellation (reaper-era optimization), \`update()\` impl. \`update()\` remains a staged stub — the interface's "rejects when state !== 'active'" clause lands with the real impl.
- **Not yet wired:** consumers of \`activate()\` never observe a quiescing handle because \`server.ts\` does not yet call \`quiesce()\`. That wiring is the reaper bead under \`ccsc-xa3.14\`.

Closes bead \`ccsc-xa3.3\`.

## Test plan

- [x] \`bun test\` — 219 pass / 0 fail / 436 expect() (up from 210)
- [x] 9 new tests: unknown-key no-op, state-transition visibility, log shape, drain-awaits-inflight, idempotent-concurrent, idempotent-post-drain, unknown-requestId no-op, duplicate-requestId rejection, quiesce-does-not-abort

Jeremy made me do it
-claude

### DIFF
--- a/server.test.ts
+++ b/server.test.ts
@@ -2585,9 +2585,8 @@ describe('createSessionSupervisor.activate', () => {
     expect(handle.state).toBe('active')
   })
 
-  test('quiesce/deactivate/shutdown are staged stubs that reject with bead pointers', async () => {
+  test('deactivate/shutdown are staged stubs that reject with bead pointers', async () => {
     const sup = makeSupervisor()
-    await expect(sup.quiesce(key)).rejects.toThrow(/xa3\.3/)
     await expect(sup.deactivate(key)).rejects.toThrow(/xa3\.14/)
     await expect(sup.shutdown()).rejects.toThrow(/xa3\.14/)
   })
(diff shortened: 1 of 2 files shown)


jeremylongshore:
Addressed the 🟢 debuggability suggestion in 2f61a76 — error now carries the SessionKey.

PR #69, opened 2026-04-19
feat(journal): SHA-256 hash-chained JournalWriter (ccsc-5pi.2, Epic 30-A)

jeremylongshore:
## Summary

Second bead under Epic 30-A. Tamper-evident append-only writer per [\`000-docs/audit-journal-architecture.md\`](000-docs/audit-journal-architecture.md) §76-161.

### Surface

- \`JournalWriter.open({ path, initialPrevHash?, now? })\` — async factory. Opens append-mode, mode \`0o600\` on create, recovers \`lastHash\` + \`nextSeq\` from the existing file's last line, registers path in module-level set to enforce single-writer-per-process. Refuses to start a new chain if the existing last line doesn't parse.
- \`writer.writeEvent(input)\` — fills in \`v\`/\`ts\`/\`seq\`/\`prevHash\`/\`hash\`, runs through \`JournalEvent.parse\` at the boundary, computes \`hash = sha256(lastHash || canonicalJson(partial))\`, appends one newline-delimited JSON line.
- Concurrent writes serialize through an internal promise queue: five concurrent \`writeEvent()\` calls yield contiguous seq 1..5 with each event's \`prevHash\` matching the previous hash.
- Fail-loud: write I/O error → broken state → subsequent writes reject. Schema failure is pre-flight — nothing lands on disk and \`nextSeq\` doesn't advance.
- \`close()\` idempotent; releases registry slot. Accessors: \`headHash\`, \`nextSequenceNumber\`.

### Helpers also landed

- \`canonicalJson()\` — RFC 8785 subset for integer-only event shapes. Sorted keys, recursive, zero whitespace. Throws on non-integer numbers / undefined / functions / symbols / bigints — a loud failure at hash time rather than silent hash divergence.
- \`sha256Hex()\` — shared so writer and verifier (ccsc-5pi.9) produce bit-identical output.
- \`WriteInput\` type — caller's payload shape.

### Scope cuts (sibling beads)

- Redaction → \`ccsc-5pi.4\` (caller responsible until then).
- Truncation → \`ccsc-5pi.6\`. CLI flag / env → \`ccsc-5pi.14\`. Verify command → \`ccsc-5pi.9\`. Rotation / reload → Epic 30-B.

Closes bead \`ccsc-5pi.2\`.

## Test plan

- [x] \`bun test\` — 251 pass / 0 fail / 546 expect() (up from 232)
- [x] 19 new tests: canonicalJson primitives / key-sort / nesting / whitespace / non-integer + unsupported-type rejection; sha256Hex known-value + shape; writer: first-write seeding, file mode 0o600, 3-event hash chain, on-disk round-trip, hash reproducibility using shared helpers, reopen recovery, reopen rejection on malformed last line, concurrent-write serialization, single-writer-per-path invariant, idempotent close + post-close rejection, schema rejection without on-disk side effects.

Jeremy made me do it
-claude

### DIFF
--- a/journal.ts
+++ b/journal.ts
@@ -36,6 +36,9 @@
  */
 
 import { z } from 'zod'
+import { createHash, randomBytes } from 'crypto'
+import { open as fsOpen, type FileHandle } from 'fs/promises'
+import { existsSync, readFileSync } from 'fs'
 import type { SessionKey } from './lib'
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


jeremylongshore:
🟡 Thread addressed in ccsc-otd — reverse-chunk-read follow-up. Commit bf1fe98... inline TODO here; actual chunked impl lands separately so this PR stays scoped to the hash-chain primitive.

PR #86, opened 2026-04-19
fix(sendable): assertSendable state-root denylist (S1)

jeremylongshore:
## Summary

`CLAUDE.md` and [`000-docs/THREAT-MODEL.md`](../blob/main/000-docs/THREAT-MODEL.md) declare that files under the state directory (`.env`, `access.json`, `audit.log`, `sessions/`) must never be sendable through the Slack upload path. The code did not enforce this — `assertSendable()` only checked an allowlist plus a basename/parent denylist. An operator who set `SLACK_SENDABLE_ROOTS` to any ancestor of `~/.claude/channels/slack` (e.g. `~/.claude`) could exfiltrate `access.json` and `audit.log` through the `reply` tool. The `.env` regex happened to catch that one bare filename by basename; nothing else in the state dir was protected.

## Fix

`assertSendable()` in `lib.ts` gains an optional `stateRoot` parameter. When supplied, we realpath-resolve both the file and the state root and reject whenever the file resolves under the state root (unless it's under the inbox, which is explicitly sendable). The state-dir check runs **before** the allowlist check so an operator-supplied allowlist cannot override the state-root denylist.

`server.ts` threads `STATE_DIR` through the `libAssertSendable` wrapper. There is one wrapper and one real call site (the `reply` tool handler); both are now guarded.

## Threat model pointer

Covers exfiltration of state-directory artifacts that `000-docs/THREAT-MODEL.md` lists under the file-upload surface — aligns code with the doc contract called out in CLAUDE.md ("when a design doc and code disagree, the code is wrong").

## New tests

Added under `describe('assertSendable — state-root denylist', ...)` in `server.test.ts`:

- `blocks access.json when operator allowlists an ancestor of the state dir` — direct misconfig, the exact exploit described above.
- `blocks a file deep inside sessions/<channel>/<thread>.json` — per-thread session files.
- `blocks a symlink OUTSIDE the state dir that points INTO it` — realpath flattens the link before comparison.
- `allows a sibling directory next to the state dir` — does not over-block.
- `legacy three-arg call (no stateRoot) preserves existing behavior` — existing callers keep passing.

## Verification

- `bun run typecheck` clean.
- `bun test` — 375 pass / 0 fail (baseline ~370, +5 new).
- `grep -n "assertSendable(" server.ts` shows both call sites; the `libAssertSendable` wrapper passes `STATE_DIR` as the fourth argument.

## Test plan

- [x] `bun run typecheck`
- [x] `bun test`
- [ ] CI Typecheck + test suite green on PR

Closes ccsc-nes

Jeremy made me do it
-claude

### DIFF
--- a/lib.ts
+++ b/lib.ts
@@ -775,6 +775,17 @@ export function assertSendable(
     throw new Error('Blocked: file does not exist or is not accessible')
   }
 
+  // Resolve the inbox once — used by both the state-dir denylist (to carve
+  // the inbox out of the state-root block) and the allowlist-roots check
+  // below.
+  const inboxReal = (() => {
+    try {
+      return realpathSync(resolve(inboxDir))
+    } catch {
+      return resolve(inboxDir)
+    }
+  })()
+
   // (1) State-dir denylist (S1). Production callers thread STATE_DIR through
   // from `server.ts`. This check runs BEFORE the allowlist so a state-dir
   // path that also happens to live under SLACK_SENDABLE_ROOTS is still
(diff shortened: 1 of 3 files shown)


jeremylongshore:
Addressed Gemini feedback — hoisted inboxReal, one resolution now serves both blocks. Test count unchanged (375 pass).

PR #87, opened 2026-04-19
fix(session): Zod schema validation in loadSession (S4)

jeremylongshore:
Part of the v0.5.1 Batch 1 multi-agent orchestration — lib.ts also touched by PR #86 (S1) in disjoint function `assertSendable`.

## Summary

- `loadSession` previously returned `JSON.parse(raw) as Session` — a cast, not validation. A corrupt or tampered session file with wrong types (e.g. `ownerId: 42`) passed load silently and reached the supervisor, which trusts `ownerId: string` for audit attribution.
- Added `SessionSchema` (Zod, `.strict()`) mirroring the `Session` interface and exported it. Unknown top-level keys are rejected — the writer controls this file; anything unexpected is treated as corruption.
- Updated the doc comment that previously admitted "does NOT schema-validate" to reflect the new reality: ZodError propagates, supervisor Quarantines the key.

## Trust-boundary rationale

The session file is written by `saveSession` (trusted) but lives on disk at a path guarded only by realpath containment. An attacker who can write to the state directory can corrupt the file with wrong types. Without schema validation, `ownerId: 42` or a missing `channel` field would silently reach the supervisor as a partially-typed `Session`. With `SessionSchema.parse()`, any deviation from the contract throws before the supervisor sees the data.

## Tests

Five new tests in the `loadSession` describe block (`server.test.ts`):

1. **Valid round-trip** — `saveSession` → `loadSession` returns identical content (confirms `SessionSchema` accepts the canonical shape).
2. **Corrupt-type rejection** — `ownerId: 42` (number) throws.
3. **Missing required field** — `key.channel` omitted throws.
4. **Unknown top-level key** — extra field throws (`.strict()` enforced).
5. **Non-JSON bytes** — garbage input throws (`JSON.parse` fires before Zod).

Test delta: 370 → 375 (5 new, 0 failures).

Closes ccsc-6j2

Jeremy made me do it
-claude

### DIFF
--- a/lib.ts
+++ b/lib.ts
@@ -11,6 +11,7 @@
 import { resolve, sep, basename, join } from 'path'
 import { realpathSync, mkdirSync, readdirSync, readFileSync, statSync, existsSync } from 'fs'
 import { writeFile, chmod, rename, unlink, readFile } from 'fs/promises'
+import { z } from 'zod'
 
 // ---------------------------------------------------------------------------
 // Constants (re-exported so server.ts and tests share the same values)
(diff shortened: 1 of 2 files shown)


jeremylongshore:
Addressed Gemini feedback — S4 tests now use async writeFile matching the async/await pattern. Test count unchanged (375 pass).

PR #90, opened 2026-04-19
fix(supervisor): quarantine state survives deactivate (S6)

jeremylongshore:
Part of the v0.5.1 Batch 1 multi-agent orchestration.

## Bug (S6)

`deactivate()` marked the handle `quarantined` then called `live.delete(id)` — the in-process quarantine signal was lost. A subsequent `activate()` re-read the session file from disk as if the save failure never happened, silently bypassing the Quarantined sticky state mandated by `000-docs/session-state-machine.md §129`.

## Fix

- `private readonly quarantined = new Map<string, Error>()` — tracks all quarantined keys with the original `Error` that caused the quarantine.
- `doActivate()` (load-failure path): records the key in `quarantined` before throwing so load failures are also sticky.
- `deactivate()` (save-failure path): calls `quarantined.set(id, err)` **before** `live.delete(id)`, ensuring no window where both maps simultaneously lack the key.
- `activate()`: rejects immediately if the key is in `quarantined`, with message `SessionSupervisor.activate: key is quarantined: <original message>`.
- `clearQuarantine(key)`: explicit operator-action unblock, aligned with the doc's "only a human (SO) clears this" clause at §129. The audit journal event for this call lands in Epic 32-B (out of scope here).
- `reapIdle()`: naturally skips quarantined keys — they are absent from `live` so the reaper never considers them.

## Doc alignment

`session-state-machine.md` describes the `Quarantined → Active: manual recovery` transition but does not name the programmatic entry point. `clearQuarantine()` is the code-side materialization of that transition; it does not contradict the doc. No doc edits required.

## Tests (`server.test.ts`)

Five new tests in `describe('SessionSupervisor quarantine (S6)')`:

1. Post-failure re-activate rejects with quarantine message.
2. Quarantine rejection message carries the original failure's text.
3. A different key (`KEY_B`) is unaffected by `KEY_A`'s quarantine.
4. `clearQuarantine` unblocks a subsequent `activate`.
5. Reaper regression guard — `reapIdle()` completes without error.

**Test delta: 370 → 375 (+5).**

Closes ccsc-bz1

Jeremy made me do it
-claude

### DIFF
--- a/server.test.ts
+++ b/server.test.ts
@@ -29,6 +29,7 @@ import {
   type Session,
   type SessionKey,
 } from './lib.ts'
+import { createSessionSupervisor } from './supervisor.ts'
 import {
   mkdtempSync,
   mkdirSync,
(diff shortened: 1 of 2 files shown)


jeremylongshore:
Addressed Gemini feedback — quarantine rejection now chains via Error.cause, preserving the original stack. Test updated to walk the cause chain. 375 pass.

PR #91, opened 2026-04-19
fix(mcp): per-tool Zod input schemas (S5)

jeremylongshore:
Part of the v0.5.1 Batch 1 multi-agent orchestration — server.ts also touched by PR #86 (S1); rebase may be required.

## Validation gap

The `CallToolRequestSchema` handler in `server.ts` destructured the incoming tool arguments as `Record<string, any>` and immediately used them in security-sensitive calls:

- `assertOutboundAllowed(args.chat_id, args.thread_ts)` — passes `undefined` straight through the outbound gate when `chat_id` is missing.
- `web.chat.postMessage({ channel, text, thread_ts })`, `web.reactions.add`, `web.chat.update`, `web.conversations.replies`, `web.conversations.history` — all called with unchecked fields. Wrong-type or missing fields bubble up as Slack API errors at best, silent pass-through at worst.

The outer MCP `CallToolRequestSchema` validates only the envelope (`name`, `arguments`), never the per-tool shape. This PR closes that defense-in-depth gap.

## Fix

Per-tool Zod schemas, co-located with the tools list in `server.ts`. All `.strict()` so a crafted call cannot smuggle in extra fields that might bypass later guards. Validation runs via `safeParse` at switch entry, before any tool body. Failures return a structured `{ content, isError: true }` result — never throw.

Error-message hygiene: Zod's default issue messages name paths + expected types (e.g. `"expected string, received number"`), never the actual value. A token-canary test locks that property in so a future Zod upgrade or a custom `.message()` call cannot regress it.

## Tools validated

| Tool | Required | Optional |
| --- | --- | --- |
| `reply` | `chat_id`, `text` | `thread_ts`, `files[]` |
| `react` | `chat_id`, `message_id`, `emoji` | `thread_ts` |
| `edit_message` | `chat_id`, `message_id`, `text` | `thread_ts` |
| `fetch_messages` | `channel` | `limit`, `thread_ts` |
| `download_attachment` | `chat_id`, `message_id` | `thread_ts` |
| `list_sessions` | — | — |

Each schema mirrors the existing `inputSchema` JSON Schema — required/optional match, so previously-accepted calls still pass. `thread_ts` is added as optional to the three tools whose handlers read `args.thread_ts` (react, edit_message, download_attachment) — the JSON Schema underdocumented it, but the handler has always used it.

## Tests

24 new tests in `server.test.ts` under `MCP tool input schemas (S5)`:

- `ReplyInput`: minimal valid, full valid, missing `chat_id`, missing `text`, empty `chat_id`, wrong-type `text`, unknown field, non-array `files`, non-string file element, error-message hygiene (token canary).
- `ReactInput`: valid, missing `emoji`, unknown field.
- `EditMessageInput`: valid, missing `message_id`.
- `FetchMessagesInput`: channel-only, full, missing `channel`, non-integer `limit`, non-positive `limit`.
- `DownloadAttachmentInput`: valid, missing `chat_id`.
- `ListSessionsInput`: empty object accepted, any extra field rejected.

Schemas are duplicated in the test file deliberately — importing from `server.ts` would trigger its top-level bootstrap (`loadEnv()` → `process.exit(1)` on missing `.env`). A comment in the test block calls out that the copies must stay in sync.

## Verification

- `bun run typecheck` — clean
- `bun test` — **394 pass / 0 fail** (baseline 370, delta +24)
- `grep 'args as any'` in `server.ts` — no matches remain

## Scope

- `server.ts` — tool registration (added `toolSchemas` block before `setRequestHandler(ListToolsRequestSchema)`) + handler entry (added `safeParse` gate at top of `setRequestHandler(CallToolRequestSchema)`).
- `server.test.ts` — new `MCP tool input schemas (S5)` describe block.

No changes to the dispatch shape, permission relay, access store, or any other subsystem.

Closes ccsc-60i

Jeremy made me do it
-claude

### DIFF
--- a/server.test.ts
+++ b/server.test.ts
@@ -1,4 +1,5 @@
 import { describe, test, expect, beforeAll, afterAll, beforeEach, afterEach } from 'bun:test'
+import { z } from 'zod'
 import {
   gate,
   assertSendable,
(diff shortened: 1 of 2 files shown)


jeremylongshore:
Addressed Gemini nit — added cross-reference comment pointing to the server.test.ts schema duplicate so future updates keep both in sync.

PR #92, opened 2026-04-19
feat(supervisor): SessionHandle.update() mutex-serialized state mutation (B1)

jeremylongshore:
Batch 3 PR A of 3 — unblocks the supervisor wiring (B2) and journal event emission (B3).

## Contract

`SessionHandle.update(fn: (prev: Session) => Session): Promise<void>` serialises state mutations through a per-handle promise-chain mutex (`writeQueue`). Each call appends a new link to the chain; links run sequentially in call order, preventing interleaving on concurrent callers.

**Mutex discipline.** The write queue tail is a `Promise<void>` field on `ConcreteHandle`. Each `update()` call chains `.then(async () => { … })` onto it and exposes a separate `callerPromise` that resolves/rejects based solely on that link's outcome. Links catch their own errors so a failed write does not collapse the chain for subsequent callers.

**Quiesce/quarantine rejection.** State is checked at enqueue time (early exit) and again inside the mutex link (re-check after queue wait). Quiescing/deactivating/activating handles reject with `"handle quiescing"`; quarantined handles reject with `"handle is quarantined"`.

**Atomic save.** Reuses `saveSession()` from `lib.ts` (tmp + chmod + rename pattern). No new I/O paths introduced.

**Save failure → quarantine.** On `saveSession()` throw: `_state` transitions to `quarantined`, `Error.cause` chains the original filesystem error, and `onQuarantine()` (injected by the supervisor at activate-time) updates the supervisor's `quarantined` map and removes the entry from `live`. This keeps the two maps consistent so a subsequent `activate()` rejects rather than returning a cached-but-corrupt handle.

## Design doc reference

See `000-docs/session-state-machine.md` §132–137 (quarantined transition), §210 invariant 1 (serialised writes per key), and §221–267 (supervisor contract).

No doc changes were required — the implementation follows the spec as written.

## New tests (413 → 419, +6)

Added `describe('SessionHandle.update (ccsc-9d9)')`:

1. **Basic update mutates + persists** — identity of in-memory `session` and on-disk JSON after a single update.
2. **Concurrent updates serialize** — three concurrent counter increments; final value must be 3, no lost writes.
3. **Concurrent update + deactivate** — fire update then quiesce/deactivate without awaiting; file must remain valid (parseable) regardless of race ordering.
4. **Update after quiesce rejects** — `/quiescing/i`.
5. **Update on quarantined handle rejects** — force quarantine via package-private `markQuarantined()`, then assert `/quarantined/i`.
6. **Save failure quarantines + chains cause** — chmod directory to `0o000` (same trick as S6), assert `handle.state === 'quarantined'`, `err.cause` matches `/EACCES|permission denied|EPERM/i`, and subsequent update also rejects with `/quarantined/i`.

Also updated the stale activate-suite test `'handle.update is a staged stub…'` → `'handle.update on an active handle resolves (ccsc-9d9 implemented)'` to reflect the real implementation.

`Closes ccsc-9d9`.

Jeremy made me do it
-claude

### DIFF
--- a/server.test.ts
+++ b/server.test.ts
@@ -5659,12 +5659,18 @@ describe('SessionHandle.update (ccsc-9d9)', () => {
     await handle.update((s) => ({ ...s, data: { count: 0 } }))
 
     // Fire three concurrent increments without awaiting individually.
-    const p1 = handle.update((s) => ({ ...s, data: { count: (s.data['count'] as number) + 1 } }))
-    const p2 = handle.update((s) => ({ ...s, data: { count: (s.data['count'] as number) + 1 } }))
-    const p3 = handle.update((s) => ({ ...s, data: { count: (s.data['count'] as number) + 1 } }))
+    // Use a safe increment helper that avoids the `as number` cast and
+    // defaults gracefully if the field is absent (per Gemini review).
+    const increment = (s: Session): Session => {
+      const current = typeof s.data['count'] === 'number' ? s.data['count'] : 0
+      return { ...s, data: { ...s.data, count: current + 1 } }
+    }
+    const p1 = handle.update(increment)
+    const p2 = handle.update(increment)
+    const p3 = handle.update(increment)
     await Promise.all([p1, p2, p3])
 
-    expect((handle.session.data['count'] as number)).toBe(3)
+    expect(handle.session.data['count']).toBe(3)
 
     // Disk must also reflect the final value.
     const path = sessionPath(stateRoot, KEY)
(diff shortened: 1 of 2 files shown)


jeremylongshore:
Addressed Gemini review in a784306:

- **supervisor.ts** — both the enqueue-time and in-queue non-active state checks now include the actual state in the error message: `"handle is not active (state: quiescing)"`. The word "quiescing" still appears in the message so the test regex `/quiescing/i` continues to match.
- **server.test.ts** — replaced the `as number` cast in the concurrent-update test with a typed increment helper that defaults safely when the field is absent.

PR #106, opened 2026-04-20
feat(policy): Epic 30-B.2 — pre-execution audit receipts

jeremylongshore:
## Summary

Post a threaded `:receipt:` Block Kit context block into the originating channel whenever a tool call passes policy (auto_allow or quorum-approved) and the channel opts in via `ChannelPolicy.audit` (`'compact'` or `'full'`). The receipt carries a correlation ID that `ccsc-dhh.3` (post-execution edit hook) will use to locate the message for the outcome update.

## What ships

**Pure helpers in `lib.ts`:**
- `generateCorrelationId()` — 8-char base64url from `crypto.randomBytes(6)`.
- `shouldPostAuditReceipt(policy)` — true only for `'compact'`/`'full'`.
- `buildAuditReceiptMessage(tool, cid)` — Block Kit context block with escaped tool name + cid.
- `escMrkdwn()` relocated from `server.ts` to `lib.ts` — was already pure; now shared between the policy-deny notice and the audit receipt.

**`server.ts` wiring:**
- Module-level `auditReceipts: Map<cid, {channel, thread, ts, tool, postedAt}>`. Populated here, consumed by `ccsc-dhh.3`.
- `postAuditReceiptIfEnabled()` — guards on `shouldPostAuditReceipt`, catches every Slack error so projection failure **never** blocks tool execution (fundamental invariant; `ccsc-dhh.6` adds backoff/reaper on top).
- Wired at three pre-execution points:
  1. `auto_allow` branch in the permission-relay — also stamps `correlationId` onto the `policy.allow` journal event.
  2. Button-handler post-quorum-approval, before the MCP allow response.
  3. Text-reply-handler post-quorum-approval, same contract.
- `PendingPermissionEntry` gains `channel` + `thread` so the receipt posts in the *originating* thread even if the approver clicks from a different UI surface.

## What this PR does NOT do

- No `correlationId` on `policy.approved` journal events — `processApprovalVote` writes the journal internally, and threading the cid through without duplicating the DRY'd helper is a separate refactor (fold into `ccsc-dhh.3` where the cid-to-journal linkage becomes load-bearing).
- No post-execution edit — that's `ccsc-dhh.3`.
- No compact/full mode differentiation — the pre-exec stub is mode-agnostic; the content of the *edit* is where mode matters (`ccsc-dhh.4`/`ccsc-dhh.5`).
- No graceful-failure reaper for orphaned receipts — `ccsc-dhh.6` territory. Current code is log-only on Slack errors.

## Invariants preserved

- **Projection never blocks execution** — all `postAuditReceiptIfEnabled` paths catch and log, never throw.
- **Default-safe** — absent `audit` field = no projection, matching every other opt-in field in this codebase.
- **mrkdwn injection defense** — tool name and cid both run through `escMrkdwn` before inclusion in the context block. Covered by dedicated tests.
- **Authoritative record unchanged** — the hash-chained local journal at `~/.claude/channels/slack/audit.log` remains the source of truth per [`000-docs/audit-journal-architecture.md`](000-docs/audit-journal-architecture.md); this PR only adds a *projection* surface.

## Test plan

- [x] `bun run typecheck` — clean.
- [x] `bun test` — **483 pass** (up from 471; +12 for the new helpers).
- [ ] Merge after Gemini review.

Closes `ccsc-dhh.2`.

Jeremy made me do it
-claude

### DIFF
--- a/lib.ts
+++ b/lib.ts
@@ -11,6 +11,7 @@
 import { resolve, sep, basename, join } from 'path'
 import { realpathSync, mkdirSync, readdirSync, readFileSync, statSync, existsSync } from 'fs'
 import { writeFile, chmod, rename, unlink, readFile } from 'fs/promises'
+import { randomBytes } from 'crypto'
 import { z } from 'zod'
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


jeremylongshore:
Addressed Gemini review on 4b067e5:

- **🟢 correlationId in log** (server.ts:354) — done in ffc2c40, resolved.
- **🟡 `as never` type bypass** (server.ts:344) — replaced with a proper `AuditReceiptContextBlock` interface in `lib.ts` mirroring Slack's context-block shape. Structural typing lets the Slack SDK accept it without importing `@slack/web-api` into `lib.ts` (which would re-couple the pure module). The call site now reads just `blocks` with no assertion. Resolved.
- **🟢 backtick escaping** (lib.ts:1532) — this is standard JS template-literal syntax: `` \` `` produces a literal backtick inside the string. Output is `` `Bash` `` → inline code in Slack. The diff viewer is showing the escape, not a literal backslash in output. Verified by the `escapes mrkdwn in attacker-controlled tool names` test which checks the rendered string. No change. Resolved.

Typecheck clean, 483 tests pass.

Jeremy made me do it
-claude

jeremylongshore:
Follow-up on Gemini's persistent backtick flag (now twice): verified the rendered output with a one-off run of \`buildAuditReceiptMessage('Bash', 'abc123xy')\`:

\`\`\`
{
  \"text\": \":receipt: audit: Bash (abc123xy)\",
  \"blocks\": [{
    \"type\": \"context\",
    \"elements\": [{
      \"type\": \"mrkdwn\",
      \"text\": \":receipt: \\\`Bash\\\` • cid \\\`abc123xy\\\`\"
    }]
  }]
}
\`\`\`

The stored string contains literal backticks (no backslashes). What Gemini sees as \`\\\\\` in the diff / JSON is the escape representation of a JS template-literal's \`\\\` → literal-backtick. Slack renders it as inline code.

Jeremy made me do it
-claude

PR #110, opened 2026-04-20
feat(lib): cap auditReceipts map at 500 entries (memory safety)

jeremylongshore:
## Summary

Close a real memory leak in the pre-execution audit receipt path shipped in #106: `auditReceipts` grows unbounded because there's no MCP tool-completion signal to trigger removal (the post-execution finalization is deferred to `ccsc-4nm` — see rationale below).

## Fix

- `AUDIT_RECEIPTS_MAX = 500` (lib.ts constant, exported).
- `enforceAuditReceiptCap(receipts, max)` — pure FIFO eviction helper. JavaScript Map preserves insertion order, so `keys().next().value` is always the oldest entry. Returns the eviction count for test observation.
- `server.ts` calls it after every `auditReceipts.set(...)` so the cap is maintained as a Map invariant, not via a separate sweep.

## Why silent eviction is correct

The **authoritative** record is the hash-chained journal. The auditReceipts map exists only to let the post-execution edit hook (future `ccsc-4nm`) find the receipt ts to update. Evicting an entry means the bridge forgets one Slack-side ts; the journal entry remains intact. Operators investigating a specific tool call consult the journal regardless — projection loss is by design.

A server handling 500+ concurrent *unresolved* pre-exec receipts either has traffic so high the thread is unreadable anyway, or has had post-exec finalization wired by then to drain entries as tools complete.

## Why the post-exec edit is deferred (`ccsc-4nm`)

The original `dhh.3` scope was \"chat.update the receipt with exit code, path, summary.\" MCP's permission-relay protocol has no tool-completion notification: we send `permissions/response allow`, Claude executes in its own process, no outcome surfaces back. Synthesizing an outcome via timer heuristics or next-action proxy would produce receipts *less* trustworthy than the journal they project. Right answer: pre-exec receipt is the projection; outcome lives in the journal. File `ccsc-4nm` for when MCP adds a completion signal.

## Tests

+6 covering the cap invariants: 500 tests total, up from 494.

- Under cap → no-op, returns 0.
- Exactly at cap → no-op.
- Over cap → evicts oldest FIFO until size equals cap, returns correct eviction count, surviving keys are the last N inserted.
- Cap of 0 → evicts everything.
- Cap larger than current size → no-op.
- Sanity bound on `AUDIT_RECEIPTS_MAX` (100 ≤ max ≤ 10k) to catch accidental regression.

## Test plan

- [x] `bun run typecheck` — clean.
- [x] `bun test` — 500 pass, 0 fail.
- [ ] Merge after Gemini.

Jeremy made me do it
-claude

### DIFF
--- a/lib.ts
+++ b/lib.ts
@@ -1612,7 +1612,7 @@ export function enforceAuditReceiptCap<V>(receipts: Map<string, V>, max: number)
     const oldestKey = receipts.keys().next().value
     if (oldestKey === undefined) break
     receipts.delete(oldestKey)
-    evicted += 1
+    evicted++
   }
   return evicted
 }
--- a/lib.ts
+++ b/lib.ts
@@ -22,6 +22,16 @@ export const MAX_PENDING = 3
 export const MAX_PAIRING_REPLIES = 2
 export const PAIRING_EXPIRY_MS = 60 * 60 * 1000 // 1 hour
 
+/** Upper bound on the in-memory auditReceipts map in server.ts. Pre-
+ *  execution receipts accumulate entries that would otherwise grow
+ *  unbounded on a long-running server — there's no MCP tool-
+ *  completion signal to clear them (see ccsc-4nm). When the map
+ *  exceeds this cap, the oldest entries are evicted FIFO via
+ *  `enforceAuditReceiptCap`. Eviction is silent (projection is
+ *  best-effort; the authoritative hash-chained journal already
+ *  captured the decision). */
+export const AUDIT_RECEIPTS_MAX = 500
+
 /** Matches permission relay replies (e.g. "y abcde", "no xyzwq").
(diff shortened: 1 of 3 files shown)


jeremylongshore:
Re: the negative-max infinite-loop flag — the current code is already safe. Trace: if \`max = -1\` and the map has entries, the while-condition is satisfied, we enter, get \`oldestKey\`, delete, repeat. When the map empties, \`receipts.keys().next().value\` returns \`undefined\`, the break fires, loop exits. No infinite loop.

Clamping \`max\` to a minimum of 0 defends against a caller this module doesn't have — the single call site passes the exported \`AUDIT_RECEIPTS_MAX\` constant (500). Per the project's CLAUDE.md rule against validating scenarios that can't happen at internal boundaries, skipping the clamp.

Resolved.

Jeremy made me do it
-claude

PR #111, opened 2026-04-20
feat(policy): lock in 31-A.4 invariant — manifest never reaches evaluate()

jeremylongshore:
## Summary

- Adds a module-level invariant doc block to `policy.ts` declaring that manifest data (from peer-bot advertisements) MUST NEVER be passed into `evaluate()`.
- Adds an import-graph lint test in `server.test.ts` that parses every import/require/re-export specifier in `policy.ts` and fails CI if any references a manifest module.
- Ships a sanity test proving the parser catches every specifier shape without false-positiving on prose in comments.

Miller 2006: advertisements are not grants. A peer's manifest is content, never authority. Role truth lives in `access.json` and `allowBotIds` — nowhere else.

No manifest module exists yet; Epic 31-A ships conditionally on a signing primitive. This bead establishes the guard *before* any 31-A code lands, so the first violation would be blocked at merge rather than discovered in production.

Closes bead `ccsc-s53.4` (Epic 31-A).

## Design references

- [`000-docs/bot-manifest-protocol.md`](./000-docs/bot-manifest-protocol.md) §91-109 — "The binding invariant"
- [`CLAUDE.md`](./CLAUDE.md) §Security Architecture — five defense layers

## Test plan

- [x] `bun test` — 502/502 pass (up from 500)
- [x] `bun test -t "31-A.4 invariant"` — 2/2 pass
- [x] Sanity: the parser catches `import … from`, `import()`, `require()`, re-exports, case variants, and ignores manifest mentions inside comments

Jeremy made me do it
-claude

### DIFF
--- a/policy.ts
+++ b/policy.ts
@@ -264,6 +264,36 @@ export function pathMatchesPrefix(resolvedPath: string, resolvedPrefix: string):
 }
 
 // ---------------------------------------------------------------------------
+// 31-A.4 Invariant — manifest data NEVER passed to evaluate()
+//
+// Design: 000-docs/bot-manifest-protocol.md §91-109 ("The binding invariant").
+//
+// Peer-bot manifests are advertisements, not grants (Miller 2006). Policy
+// decisions MUST use only verified signals carried on ToolCall (tool,
+// sessionKey.channel, actor) — never content a peer advertised about itself.
+// A peer claiming "I am an approver" in its manifest does not make it one;
+// allowBotIds and access.json are the only sources of role truth.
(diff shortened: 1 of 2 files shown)


jeremylongshore:
Thanks for the review. Addressed the one stylistic call-out in ae02f1b: `extractImportSpecifiers` is now at module scope in `server.test.ts` alongside the existing helpers (`makeAccess`, `makeOpts`). Typecheck clean, 502/502 tests still pass.

Jeremy made me do it
-claude

PR #112, opened 2026-04-20
feat(lib): add assertPublishAllowed gate — only access.allowFrom may publish

jeremylongshore:
## Summary

Epic 31-B.5 (bead `ccsc-0qk.5`). Lands the publish-manifest authorization gate ahead of the `publish_manifest` MCP tool itself, symmetric to the 31-A.4 "manifest never reaches `evaluate()`" invariant on the read side.

- **`assertPublishAllowed(ownerId, access)`** in `lib.ts` — throws on any caller not in `access.allowFrom`.
- One allowlist governs both DM access AND publish authority. Operators have a single consistent place to manage who can act on behalf of the bot.
- Default-safe: `defaultAccess()` returns `allowFrom: []` so nobody can publish until an operator explicitly adds a user.

## Test coverage (5 new tests, 507/507 pass total)

- user in `allowFrom` → passes
- user not in `allowFrom` → throws with an ID- and field-bearing reason
- empty `allowFrom` rejects everyone (fail-closed default locked in)
- case-sensitive exact match (Slack IDs are opaque and exact-match)
- per-channel `ChannelPolicy.allowFrom` is deliberately NOT consulted — channel-level inbound allowlists cannot accidentally grant publish authority

Closes bead `ccsc-0qk.5`.

## Test plan

- [x] `bun test` — 507/507 pass
- [x] `bun test -t "assertPublishAllowed"` — 5/5 pass

Jeremy made me do it
-claude

### DIFF
--- a/lib.ts
+++ b/lib.ts
@@ -976,6 +976,34 @@ export function assertOutboundAllowed(
 }
 
 /**
+ * Assert that a given user_id may publish a manifest (Epic 31-B.5).
+ *
+ * `publish_manifest` is an outbound act performed on behalf of the session
+ * owner. Only users in `access.allowFrom` — the top-level DM allowlist —
+ * may publish, giving operators one consistent authorization surface:
+ * the same list that gates DMs gates publishing.
+ *
+ * Throws on rejection so MCP tool handlers surface a clear failure to
+ * Claude and, via journal, to the operator. Default-safe: `allowFrom` of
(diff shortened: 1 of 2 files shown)


jeremylongshore:
Addressed in c9e7691 — single template literal, message content unchanged so the `/access\.allowFrom/` test still matches character-for-character. Typecheck clean, 507/507 tests pass.

Jeremy made me do it
-claude

PR #114, opened 2026-04-20
feat(manifest): add read_peer_manifests MCP tool

jeremylongshore:
## Summary

Epic 31-A.2 (bead \`ccsc-s53.2\`). Lands the read surface that returns validated v1 peer-bot manifests to Claude as tool text. Design: [\`000-docs/bot-manifest-protocol.md\`](./000-docs/bot-manifest-protocol.md) §58-87.

Split between the pure and the wired:

- **\`manifest.ts\`** — \`extractManifests(texts)\`: cheap string-includes pre-filter on the magic header key; silent drop on JSON parse failure, Zod failure, or non-string bodies; position-preserving, non-deduping. Size cap deliberately out of scope (sibling bead \`ccsc-s53.3\`).
- **\`server.ts\`** — \`read_peer_manifests\` MCP tool. Outbound-gate reuse gates read access (no new surface into non-opted channels). Fetches pinned messages AND the last 50 channel messages per the doc §79, passes to \`extractManifests\`, returns JSON with a stable \`{channel, count, manifests[]}\` shape. Either fetch failing falls through cleanly — this is advertising, not a transaction.

## Silent-drop posture

Per §81 of the protocol doc: malformed JSON, failed Zod validation, missing magic header, truthy-but-not-\`true\` magic values, non-string bodies — all dropped without surfacing to callers. There is no per-message error channel because manifests are *advertising*, not an API.

## Test coverage (10 new tests, 533/533 total pass, up from 523)

- Empty input → []
- No magic header anywhere → []
- Single valid → extracted
- Mixed valid/invalid preserves input order
- Malformed JSON dropped
- Zod-failing payload dropped (non-SemVer version)
- **Magic-header truthy-imposters** (\`'yes'\`, \`1\`) dropped — binding invariant preserved
- \`null\`, \`undefined\`, empty-string bodies skipped safely
- Pre-filter drops JSON that never mentions the magic key
- Explicit no-dedup contract when a manifest appears in both pins and history

## Test plan

- [x] \`bun test\` — 533/533 pass
- [x] \`bun test -t "extractManifests"\` — 10/10 pass
- [x] Tool registered in both \`toolSchemas\` and the \`ListTools\` response

Size cap (\`ccsc-s53.3\`), rate limit + 5-min cache (\`ccsc-s53.5\`), and formal malformed-drop tests (\`ccsc-s53.13\`) layer on in sibling beads without changing this surface.

Closes bead \`ccsc-s53.2\`.

Jeremy made me do it
-claude

### DIFF
--- a/manifest.ts
+++ b/manifest.ts
@@ -142,19 +142,15 @@ function looksLikeManifest(text: string): boolean {
 export function extractManifests(
   texts: ReadonlyArray<string | null | undefined>,
 ): ManifestV1[] {
-  const out: ManifestV1[] = []
-  for (const text of texts) {
-    if (typeof text !== 'string' || text.length === 0) continue
-    if (!looksLikeManifest(text)) continue
-    let parsed: unknown
+  return texts.flatMap((text) => {
+    if (typeof text !== 'string' || text.length === 0) return []
+    if (!looksLikeManifest(text)) return []
     try {
-      parsed = JSON.parse(text)
+      const parsed: unknown = JSON.parse(text)
+      const result = ManifestV1.safeParse(parsed)
+      return result.success ? [result.data] : []
     } catch {
-      continue
+      return []
     }
-    const result = ManifestV1.safeParse(parsed)
-    if (!result.success) continue
-    out.push(result.data)
-  }
-  return out
+  })
 }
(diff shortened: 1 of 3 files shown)


jeremylongshore:
Both suggestions addressed in 9909146:

- \`extractManifests\` now uses \`flatMap\` — same silent-drop behavior, cleaner shape.
- \`read_peer_manifests\` now fetches pins + history concurrently via \`Promise.allSettled\` — latency drops from (pins + history) to max(pins, history). Failure posture unchanged.

Typecheck clean, 533/533 tests still pass.

Jeremy made me do it
-claude

PR #115, opened 2026-04-20
feat(manifest): 40 KB raw-body size cap before JSON.parse

jeremylongshore:
## Summary

Epic 31-A.3 (bead \`ccsc-s53.3\`). Bounds the cost of \`JSON.parse\` on a single peer-posted manifest so a hostile pinned message cannot force gigabyte-scale allocation. Design: [\`000-docs/bot-manifest-protocol.md\`](./000-docs/bot-manifest-protocol.md) §47-49.

- **\`MAX_MANIFEST_BYTES = 40 * 1024\`** — exported constant so tests and operators reference the same number.
- **UTF-8 byte count via \`TextEncoder\`** — not \`string.length\`, which would count UTF-16 code units and under-report for multi-byte code points. A 4-byte emoji is 4 bytes, not 2.
- **Layered after pre-filter, before \`JSON.parse\`** — only candidates pay the byte-count walk, and the parser never sees oversized input.
- **Oversize drops log \`console.debug\`** — malformed-JSON and Zod-failure drops stay silent per §81. Asymmetry is deliberate: oversize is operator-actionable (possible DoS); the others are noise.

## Test coverage (3 new tests, 536/536 total pass)

| Test | What it proves |
|---|---|
| At cap (40960 bytes, whitespace-padded) | Boundary is inclusive; whitespace-padded JSON still parses + validates |
| One byte over (40961) | Silent drop; sanity asserts the payload WOULD be a valid manifest if it got past the cap, isolating the cap as the cause; debug line verified to name "oversized" + actual byte count |
| Multi-byte (🔥 padding) | Cap measured in bytes, not characters — totalBytes > cap while totalString.length < totalBytes; still rejects |

## Test plan

- [x] \`bun test\` — 536/536 pass
- [x] \`bun test -t "extractManifests"\` — 13/13 pass

Closes bead \`ccsc-s53.3\`.

Jeremy made me do it
-claude

### DIFF
--- a/manifest.ts
+++ b/manifest.ts
@@ -95,6 +95,19 @@ export type ManifestV1 = z.infer<typeof ManifestV1>
  */
 export const MANIFEST_V1_MAGIC_KEY = '__claude_bot_manifest_v1__' as const
 
+/**
+ * Hard cap on a single manifest's raw body, in bytes after UTF-8 encode
+ * (Epic 31-A.3). The doc (§47) phrases the cap as "≤ 40 KB"; we treat
+ * "KB" as 1024 bytes, the convention used everywhere else in this
+ * codebase for memory-safety constants. Anything over is silently
+ * dropped before `JSON.parse` is called — the point of the cap is to
+ * bound the cost of the parse step, which can blow up memory
+ * super-linearly on deeply-nested or hostile payloads. A peer that can
+ * make us allocate 100 MB to parse one pinned message is a cheap DoS
+ * otherwise.
+ */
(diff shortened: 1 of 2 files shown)


jeremylongshore:
Addressed in d58ffd1 — multi-byte oversize test now has the same \`debugCalls\` capture + \`/oversized/i\` + byte-count assertions as the single-byte test above it. Both oversize tests now pin the diagnostic-message contract. (The 'Test comment to check if commenting works' appears to be review-bot meta-noise; skipping that one.) Typecheck clean, 536/536 tests pass.

Jeremy made me do it
-claude

PR #116, opened 2026-04-20
feat(manifest): per-channel 5-min read cache + rate limit

jeremylongshore:
## Summary

Epic 31-A.5 (bead \`ccsc-s53.5\`). Caches \`read_peer_manifests\` results per-channel so repeated tool calls within the TTL window return instantly instead of re-hitting Slack. Doubles as the rate limit per [\`000-docs/bot-manifest-protocol.md\`](./000-docs/bot-manifest-protocol.md) §84: "one fetch per channel per 5 min."

## Changes

- **\`manifest.ts\`** — new \`createManifestCache({now, maxEntries, ttlMs})\` factory → \`{get, set, clear, size}\`. Injected time source for testability. Soft LRU, 256-entry default cap. Lazy eviction of expired entries on \`get\`. Exports \`MANIFEST_CACHE_TTL_MS = 5 * 60 * 1000\`.
- **\`journal.ts\`** — adds \`manifest.read\` and \`manifest.read.cached\` to \`EventKind\` (21 total, up from 19). Doc §163-166 mandates a journal entry on every read so activity stays forensically visible even when served from cache.
- **\`server.ts\`** — module-level \`manifestCache\`; tool handler hits cache first, journal \`manifest.read.cached\` on hit, existing fetch+extract path on miss + \`manifest.read\` + \`cache.set\`.

Also fixes a doc drift: the \`extractManifests\` JSDoc previously said the 40 KB cap was a sibling bead; now correctly notes the cap is enforced in that function.

## Test coverage (11 new cache tests, 547/547 total)

| Test | What it proves |
|---|---|
| \`MANIFEST_CACHE_TTL_MS\` value | 5 * 60 * 1000 ms pin |
| Miss on empty | returns undefined |
| Set → get within TTL | returns payload |
| Get at exactly \`ttlMs\` from \`cachedAt\` | expired (fail-closed boundary) |
| Expired entry lazily evicted on \`get\` | \`size\` decreases |
| \`set\` overwrites \`cachedAt\` on existing key | TTL resets |
| LRU: insert at cap | oldest \`cachedAt\` evicted |
| Overwrite at cap | no eviction |
| \`clear\` | drops everything |
| \`channelId\` isolation | no cross-channel leakage |
| Default TTL | 5 min when omitted |

Plus an update to the \`EventKind\` pin (19 → 21) that caught a doc-count drift in the first test run.

## Test plan

- [x] \`bun test\` — 547/547 pass
- [x] \`bun test -t "createManifestCache"\` — 11/11 pass

Closes bead \`ccsc-s53.5\`.

Jeremy made me do it
-claude

### DIFF
--- a/journal.ts
+++ b/journal.ts
@@ -13,7 +13,7 @@
  *     v:        1,                          // schema version; chain refuses mixed versions
  *     ts:       '2026-04-19T12:34:56.789Z', // ISO-8601 UTC with ms precision
  *     seq:      42,                         // monotonic per chain
- *     kind:     'gate.inbound.drop',        // discriminated union of 19 event kinds
+ *     kind:     'gate.inbound.drop',        // discriminated union of 21 event kinds
  *     toolName?: 'reply',
  *     input?:   { chat_id: 'C01', text: '...' },     // post-redaction
  *     outcome?: 'allow' | 'deny' | 'require' | 'drop' | 'n/a',
@@ -78,9 +78,14 @@ export const EventKind = z.enum([
   'system.boot',
   'system.shutdown',
   'system.reload',
+  // Bot-manifest protocol (Epic 31-A). Emitted on every read_peer_manifests
+  // call so manifest activity is forensically visible even when cached;
+  // see 000-docs/bot-manifest-protocol.md §163-166.
+  'manifest.read',
+  'manifest.read.cached',
 ])
 
-/** TypeScript string union of the 19 event kinds. */
+/** TypeScript string union of the 21 event kinds. */
 export type EventKind = z.infer<typeof EventKind>
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


jeremylongshore:
Addressed both in d362641:

- **DRY hit/miss**: resolved \`manifests\` + \`manifestEventKind\` into variables first, then fall through to a single \`journalWrite\` + single \`return\`. Same behavior, cleaner shape.
- **O(1) LRU**: respectfully declined. The set() comment already documents the tradeoff — n ≤ 256 (larger than any realistic workspace channel count), and each miss is bounded by a Slack round-trip that dominates by orders of magnitude. Doubly-linked-list machinery would be speculative perf work; happy to revisit if maxEntries ever needs to grow past low thousands.

Typecheck clean, 547/547 tests still pass.

Jeremy made me do it
-claude

PR #120, opened 2026-04-20
feat(manifest): 8 KB publish-side size cap

jeremylongshore:
## Summary

Epic 31-B.2 (bead \`ccsc-0qk.2\`). Adds a stricter size cap on outgoing manifests than the 40 KB read cap — Postel's Law: *be conservative in what you send, liberal in what you receive*. A publisher writes content it controls; tightening the cap catches operator mistakes (e.g. an accidentally pasted API payload) at publish time.

## Changes

- **\`manifest.ts\`** — new \`MAX_PUBLISH_MANIFEST_BYTES = 8 * 1024\` constant + \`assertPublishSizeAndSerialize(manifest)\` helper that serializes with \`JSON.stringify(..., null, 2)\` (same format as the publisher), measures UTF-8 bytes via \`TextEncoder\`, throws on over-cap with an ID-bearing message, and returns the serialized string on success so the caller can post it directly.
- **\`server.ts\`** — \`publish_manifest\` handler calls \`assertPublishSizeAndSerialize\` after auth gates and before the replace sweep. The returned string is passed to \`chat.postMessage\`. Rejection journals a \`gate.outbound.deny\`.

## "Bytes measured = bytes on the wire"

Keeping the serialize + size-check in one function guarantees there's no formatter drift between what we validate and what we post. A split signature (\`assertPublishSize(manifest)\` + separate \`JSON.stringify\` in the caller) would leave room for the cap to silently rise.

## Test coverage (7 new, 564/564 total)

| Test | What it proves |
|---|---|
| Constant value | \`MAX_PUBLISH_MANIFEST_BYTES = 8192\` |
| Postel invariant | \`< MAX_MANIFEST_BYTES\` (8 KB publish < 40 KB read) |
| Normal manifest | Passes, returns parseable JSON |
| Near-cap binary search | Cap is inclusive on the pass side |
| One byte over | Rejects with \`/Publish size/\` |
| Oversize tools[] | Message names actual bytes + cap + the fields to shrink |
| Multi-byte emoji padding | Cap measured in UTF-8 bytes, not UTF-16 units |

## Test plan

- [x] \`bun test\` — 564/564 pass
- [x] \`bun test -t "assertPublishSizeAndSerialize"\` — 7/7 pass

Closes bead \`ccsc-0qk.2\`.

Jeremy made me do it
-claude

### DIFF
--- a/manifest.ts
+++ b/manifest.ts
@@ -330,8 +330,8 @@ export function createManifestCache(
  * directly. Keeping serialization and size-check in one function
  * guarantees the bytes we measure are the bytes we post — a split
  * signature (`assertPublishSize(manifest)` + separate
- * `JSON.stringify` in the caller) would leave room for a formatter
- * drift to silently raise the effective cap.
+ * `JSON.stringify` in the caller) would leave room for formatting
+ * differences to silently raise the effective cap.
  *
  * Stricter than the 40 KB read-side cap (Postel's Law — strict on
  * output, liberal on input). See `MAX_PUBLISH_MANIFEST_BYTES` for the
--- a/manifest.ts
+++ b/manifest.ts
@@ -108,6 +108,19 @@ export const MANIFEST_V1_MAGIC_KEY = '__claude_bot_manifest_v1__' as const
  */
 export const MAX_MANIFEST_BYTES = 40 * 1024
 
+/**
+ * Publish-side cap on the serialized manifest (Epic 31-B.2, bead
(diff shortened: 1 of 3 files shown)


jeremylongshore:
Both addressed in 78ac4f7:

- "formatter drift" → "formatting differences" in the JSDoc.
- Over-cap rejection test now uses chained \`expect(...).toThrow(regex)\` instead of try/catch-and-inspect — one regex per diagnostic facet (prefix / actual bytes / cap / shrinkable-fields hint).

Typecheck clean, 564/564 tests still pass.

Jeremy made me do it
-claude

PR #121, opened 2026-04-20
feat(manifest): 1-publish-per-channel-per-hour rate limiter

jeremylongshore:
## Summary

Epic 31-B.4 (bead \`ccsc-0qk.4\`). Caps publish frequency to prevent pin-spam. In-memory only, reset on process restart — same posture as the 31-A.5 read cache.

## Changes

- **\`manifest.ts\`** — \`PUBLISH_RATE_LIMIT_MS = 60 * 60 * 1000\` + \`createPublishRateLimiter({now, windowMs, maxEntries})\` factory returning \`{checkAndRecord, clear, size}\`. Design symmetric to \`createManifestCache\`: factory, injected time source, soft LRU at 256 entries. Rejection error names the channel, how long ago the last publish was, and approximate minutes-remaining.

- **\`server.ts\`** — module-level \`publishRateLimiter\`, wired into \`publish_manifest\` handler **after auth gates + size cap, before replace sweep**. Slot reservation happens at the check, so a downstream Slack failure does NOT roll back the stamp — a retry-on-failure limiter would be trivially bypassable.

## Gate order (publish_manifest)

1. \`assertPublishAllowed\` — only \`access.allowFrom\` users
2. \`assertOutboundAllowed\` — channel opt-in
3. \`assertPublishSizeAndSerialize\` — 8 KB serialized cap
4. **\`publishRateLimiter.checkAndRecord\`** — ← this PR
5. replace sweep + post + pin + journal

## Test coverage (10 new, 574/574 total)

| Test | What it proves |
|---|---|
| Constant value | \`PUBLISH_RATE_LIMIT_MS = 3 600 000 ms\` |
| First publish | Passes, size() = 1 |
| Second within window | Throws \`/Publish rate limit/\` + channel ID + \`1-per-hour\` |
| Boundary at exactly windowMs | Window is \`[0, windowMs)\` — allowed at the boundary |
| Post-window refresh | Second publish updates stamp (next check rejects against refreshed) |
| Channel isolation | Cooldown on one channel doesn't affect another |
| \`clear()\` | Drops everything |
| Soft-LRU at cap | Oldest evicted on insert |
| Default window | 1 hour when \`opts.windowMs\` omitted |
| Rejection message | Includes \`10m ago\` + \`~50m\` remaining |

## Test plan

- [x] \`bun test\` — 574/574 pass
- [x] \`bun test -t "createPublishRateLimiter"\` — 10/10 pass

Closes bead \`ccsc-0qk.4\`.

Jeremy made me do it
-claude

### DIFF
--- a/manifest.ts
+++ b/manifest.ts
@@ -337,6 +337,103 @@ export function createManifestCache(
  * output, liberal on input). See `MAX_PUBLISH_MANIFEST_BYTES` for the
  * reasoning.
  */
+// ---------------------------------------------------------------------------
+// Publish rate limiter (Epic 31-B.4, bead ccsc-0qk.4)
+// ---------------------------------------------------------------------------
+
+/**
+ * One publish per channel per hour. Prevents pin-spam if an operator
+ * (or a misbehaving automation) triggers repeated publishes. Not
+ * persisted — a process restart clears the window, which is the
+ * accepted tradeoff for keeping this purely in-memory (doc §166 has
+ * the same "restart clears" posture for the read cache).
+ */
+export const PUBLISH_RATE_LIMIT_MS = 60 * 60 * 1000
+
+export interface PublishRateLimiter {
+  /**
+   * Check whether `channelId` is currently in a cooldown window and,
+   * if not, record `now()` as the most-recent publish timestamp. The
+   * record happens BEFORE the caller's actual publish — reserving the
+   * slot — so a concurrent or retried call sees the cooldown even if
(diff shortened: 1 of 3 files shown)


jeremylongshore:
Noted. Same call as PR #116 — the O(N) scan at 256 entries is bounded by the Slack round-trip cost that dominates by orders of magnitude, and Gemini itself notes "not a performance concern" at the default cap. Happy to revisit if maxEntries ever needs to grow past low thousands. No change.

Jeremy made me do it
-claude

PR #123, opened 2026-04-20
test(manifest): round-trip + replace-sweep coverage

jeremylongshore:
## Summary

Two Epic 31-B.16 leaves in one PR:

### ccsc-0qk.7 — round-trip publish → read-extract

Pure-function chain (\`assertPublishSizeAndSerialize → extractManifests\`) proves what the publisher writes parses back to the same manifest object the read path produces. Three cases:

1. Minimal manifest round-trips byte-for-field
2. Fully-populated manifest (every optional field including \`agentCard\`) round-trips without loss
3. 1000-char-description fixture pins the Postel-Law safety margin (publish body fits read cap with ~5× headroom)

### ccsc-0qk.10 — replace-sweep filter

Extract \`findOurPriorManifestPins\` from the \`publish_manifest\` handler into \`manifest.ts\` so it can be tested without a Slack WebClient mock. Pure signature:

\`\`\`ts
(items: PinItemLike[], identity: {botId?, botUserId?}) => string[]
\`\`\`

Filter semantics, each tested as a distinct reject path:

- Identity with no \`bot_id\` AND no \`user\` → \`[]\` (fail-closed before bootstrap populates either)
- File-kind pins ignored (not messages)
- Peer-bot manifest pins NOT touched (different \`bot_id\` + \`user\`)
- Our own non-manifest pins NOT touched (missing magic header)
- Our own messages with no \`ts\` skipped
- Multiple prior manifests all surface (replace-sweep cleanup after a crashed mid-flight publish)
- \`bot_id\` match alone or \`user\` match alone each suffice

\`server.ts publish_manifest\` handler now delegates the filter to the extracted function.

## Also closes by reference

- **ccsc-0qk.8** (rate limiter rejects 2nd within 1h) — covered by PR #121 \`createPublishRateLimiter\` tests
- **ccsc-0qk.9** (non-allowFrom caller rejected) — covered by PR #112 \`assertPublishAllowed\` tests

Completes sub-epic \`ccsc-0qk.16\`.

## Test plan

- [x] \`bun test\` — 594/594 pass (up from 581)
- [x] \`bun test -t "findOurPriorManifestPins"\` — 10/10 pass
- [x] \`bun test -t "publish → read round-trip"\` — 3/3 pass

Closes beads \`ccsc-0qk.7\` and \`ccsc-0qk.10\`.

Jeremy made me do it
-claude

### DIFF
--- a/manifest.ts
+++ b/manifest.ts
@@ -400,6 +400,79 @@ export function createManifestCache(
  * reasoning.
  */
 // ---------------------------------------------------------------------------
+// findOurPriorManifestPins — pure filter for the replace sweep (ccsc-0qk.10)
+// ---------------------------------------------------------------------------
+
+/**
+ * Shape we need from a pins.list item to decide whether it's a prior
+ * manifest we posted. Kept deliberately minimal so the consumer can
+ * pass Slack's full item shape via structural typing without a runtime
+ * coercion — every field is optional so a file-kind pin (no `.message`)
+ * is handled gracefully.
+ */
+export interface PinItemLike {
+  readonly type?: string
+  readonly message?: {
+    readonly text?: string | null
+    readonly bot_id?: string
+    readonly user?: string
+    readonly ts?: string
+  }
+}
+
+/** Identity handles needed to recognise our own posts in a pins.list
+ *  response. Either field may be empty when bootstrap hasn't finished
+ *  populating them; the filter treats empty values as "don't match". */
+export interface BotIdentity {
+  readonly botId?: string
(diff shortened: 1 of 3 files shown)


jeremylongshore:
Addressed in 15d94cf — \`PinItemLike\` is now in the named import at the top of \`server.ts\`, dropping the inline \`import('./manifest.ts').PinItemLike\` at the call-site. Typecheck clean, 594/594 tests still pass.

Jeremy made me do it
-claude

PR #144, opened 2026-04-22
feat(skill): /slack-channel:policy — author policy rules without hand-editing access.json

jeremylongshore:
## Summary

- Adds new operator skill `skills/policy/SKILL.md` with four subcommands: `list`, `lint`, `add <id> <effect> <json-match> [opts]`, `remove <id>`.
- Adds `scripts/policy-validate.ts` — thin CLI wrapper around the real `parsePolicyRules()` + `detectShadowing()` + `detectBroadAutoApprove()` from `policy.ts`, so skill-authored rules go through the same validation the server uses at boot.
- Validator also implements the **duplicate-id check** that `ACCESS.md §"Safety checks"` promises but `server.ts loadPolicy()` does not enforce today. Drift tracked in **ccsc-kx8**.
- All writes atomic (`access.json.tmp` → rename) with `chmod 0o600`.
- No server-side changes; no hot reload (intentionally). Every success message tells the operator to restart.

Closes **ccsc-0ss**.

## Design notes

- **Validator as a script, not a server flag.** Keeps `server.ts` unchanged (no new boot-time surface) while giving the skill a deterministic pre-write check. Script exit code is the contract: `0` = parse OK (with optional warnings), `1` = refuse to write.
- **Validator output is always valid JSON on stdout.** The skill parses it rather than scraping stderr, so messages can move without breaking the skill.
- **Duplicate-id check lives in the validator, not the server.** Philosophically it should be in `policy.ts` so server and validator share one code path; that's ccsc-kx8. Shipping the check here first means the skill can't silently land dupes today.
- **Roll-back on post-write validation failure.** The skill's `add` path writes, re-validates, and rolls back if the re-validation fails. This covers the "my rule was valid individually but shadows something existing" edge case.

## Test plan

- [x] `bunx @biomejs/biome check .` — clean (Biome reformatted the new script's imports on first pass; committed)
- [x] `bun test --timeout 15000` — 682 pass, 0 fail
- [x] `bash scripts/coverage-floor.sh 95` — 98.39% / 98.75%
- [x] `bunx depcruise --config .dependency-cruiser.js .` — no violations (scripts/ is excluded)
- [x] `bash scripts/gherkin-lint.sh --path features/ --strict` — 0 warnings
- [x] `bash scripts/harness-hash.sh --verify` — OK (no pinned files touched)
- [x] `bun audit --audit-level=high --ignore=GHSA-j3q9-mxjg-w52f` — clean
- [x] `bun scripts/crap-score.ts --threshold 30` — 0 over (max 28)
- [x] Smoke tests of `scripts/policy-validate.ts` (empty, valid, malformed, duplicate-id, broad-auto-approve) all return the expected exit codes + JSON shapes.
- [x] Smoke test against a realistic `access.json` with 3 mixed-effect rules — count 3, no warnings.

Jeremy made me do it
-claude

### DIFF
--- a/ACCESS.md
+++ b/ACCESS.md
@@ -284,6 +284,10 @@ A missing or empty `policy` field means "no authored rules" and is **not** an er
 
 Epic 29-B wired the loader and the `evaluate()` call into the permission-relay handler. The wiring lives in `server.ts` at the `PermissionRequestSchema` handler — see the `decidePermissionRoute()` helper in `lib.ts` for the pure decision-routing logic.
 
+### Authoring rules without hand-editing
+
+The `/slack-channel:policy` skill (`skills/policy/SKILL.md`) is the ergonomic front door to authoring rules. It wraps atomic `access.json` writes with pre-write validation via `scripts/policy-validate.ts`, which runs the same `parsePolicyRules()` + `detectShadowing()` + `detectBroadAutoApprove()` functions the server uses at boot. Subcommands: `list`, `lint`, `add <id> <effect> <json-match> [opts]`, `remove <id>`. The skill complements the hand-edit path; it does not replace it. Hot reload remains unsupported — every successful mutation ends with a "restart the server" notice.
+
 ## File attachments — sendable roots
 
 The `reply` tool can attach files to Slack messages, but only files whose
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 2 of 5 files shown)


jeremylongshore:
Addressed in 14921ff — replaced `args[0]!` / `args[1]!` non-null assertions and the `as Record<string, unknown>` cast with an early-return guard that narrows `firstArg` and `rulesPath` cleanly. Applied the same pattern to the `--rules` branch which had the same shape. Behavior unchanged (smoke tests verify exit codes: valid=0, malformed=1, no-args=1).

Jeremy made me do it
-claude

jeremylongshore:
Round 2 addressed in 8adc2f6:

- 🟡 `scripts/policy-validate.ts:59` — Replaced `as Record<string, unknown>` with a structural `'policy' in body` type guard; the narrowed cast stays inside the guard where it's safe. Behavior preserved (smoke tests: valid policy, empty object, both exit 0 with expected JSON).
- 🟢 `skills/policy/SKILL.md:7` — `argument-hint` now matches the Usage section verbatim (explicit flag list, no `[opts]` shorthand).

The two 🟢 suggestions on `CHANGELOG.md:11` and `ACCESS.md:290` were false positives — the backtick-wrapped syntax they proposed is already in the file. No changes needed there.

Jeremy made me do it
-claude

PR #146, opened 2026-04-22
fix(policy): enforce duplicate-rule-id rejection server-side (ccsc-kx8)

jeremylongshore:
## Summary

Closes **ccsc-kx8**. Fixes a doc-vs-code drift surfaced while shipping PR #144 (policy skill).

`ACCESS.md` §"Safety checks the loader runs" has always promised duplicate-id rejection is **fatal at boot**, but `server.ts loadPolicy()` never wired the check — only the validator script shipped alongside `/slack-channel:policy` caught it. A policy with two rules sharing an `id` would load cleanly at boot, and the second rule would be silently unreachable because `evaluate()` uses first-applicable combining.

## Changes

- **`policy.ts`** — new `assertUniqueRuleIds()` exported as a sibling to `detectShadowing` and `detectBroadAutoApprove`. Throws (does not warn) — matches the documented fatal contract. Error message enumerates **every** duplicated id, sorted and deduped, so an operator fixing a large policy sees the full set in one shot. Doc comment on `parsePolicyRules` updated to reference the new sibling instead of just saying "NOT enforced here".
- **`server.ts loadPolicy()`** — runs `assertUniqueRuleIds()` immediately after `parsePolicyRules()` succeeds. Fails boot with a descriptive error (`[slack] policy rule-id uniqueness failed at boot: …`), same fail-closed shape as the existing parse-error path.
- **`scripts/policy-validate.ts`** — swapped its local `findDuplicateIds` helper for the shared primitive. Skill and server now use **one code path**.
- **`server.test.ts`** — 4 new tests covering empty input, all-unique ids, single duplicate, and multi-duplicate-sorted-dedup enumeration.

## Design rationale

Kept uniqueness separate from `parsePolicyRules()` (which still passes duplicates through — locked in by an existing test at line 2594). Two reasons:
1. Lets the loader report parse errors and uniqueness errors **independently** — if both are wrong, the operator sees the schema error first (fix that, boot again, see the uniqueness error).
2. Lets pure-parse callers (e.g. dry-run tools) opt out of the throw contract.

Throws rather than returns warnings (like `detectShadowing` does) because shadowing is operator-intentional-sometimes while duplicate ids are never correct.

## Test plan

- [x] `bunx @biomejs/biome check .` — clean
- [x] `bun test --timeout 15000` — **686 pass / 0 fail / 4,009 expects** (was 682 / 4,005)
- [x] `bash scripts/coverage-floor.sh 95` — **98.46% line / 98.82% func** (bumped up from 98.39 / 98.75)
- [x] `bunx depcruise --config .dependency-cruiser.js .` — no violations
- [x] `bash scripts/harness-hash.sh --verify` — OK
- [x] `bun scripts/crap-score.ts --threshold 30` — max 28, 0 over
- [x] `bun audit --audit-level=high --ignore=GHSA-j3q9-mxjg-w52f` — clean
- [x] Validator smoke: duplicate-id fails (ok:false with new clean message), malformed fails, valid passes.

Jeremy made me do it
-claude


jeremylongshore:
Declining the suggestion on server.ts:430 — it's a false positive.

`\\n` inside a **template literal** (backticks) is the newline escape, not a literal backslash-n. Proof: the existing parse-error branch 10 lines above (`[slack] policy parse failed at boot: ...`) uses the identical pattern and has been in production since Epic 29-B, producing clean multi-line stderr output. My new branch was written to match that sibling's style on purpose.

Gemini's alternative (single multi-line template literal) is stylistically valid but doesn't fix a bug. Keeping the current shape for consistency with the existing parse-error format.

Jeremy made me do it
-claude

PR #148, opened 2026-04-22
refactor(journal): reverse-chunk tail read + single file handle (ccsc-otd)

jeremylongshore:
## Summary

Closes **ccsc-otd**. Originally flagged in Gemini review on PR #69.

\`JournalWriter.open()\` previously \`readFileSync\`'d the entire audit log just to pull \`lastHash\` and \`nextSeq\` off the last line — **O(file size)** in both memory and startup time. Fine at MB-scale, stops scaling past that.

## Changes

- **Reverse-chunk tail scan.** New module-private \`readLastLine(fh)\` helper reads 64 KiB windows from EOF backwards until it finds the newline preceding the last non-empty line. Memory drops to **O(last line size)**. Chunk size is a named constant (\`READ_LAST_LINE_CHUNK_SIZE\`).
- **Single file handle.** Unifies the old pattern (\`readFileSync\` → separate \`fsOpen('a')\`) into **one** \`fsOpen('a+', 0o600)\` throughout. Reads use explicit \`position\` so they don't disturb the O_APPEND pointer; writes still land atomically at EOF. **Incidentally closes the stat-then-use TOCTOU window** that triggered the earlier CodeQL \`js/file-system-race\` pattern (alert #22, dismissed as out-of-threat-model yesterday but now structurally gone).
- **FD safety.** If recovery throws (invalid JSON last line, etc.), the handle is \`.close()\`d before the throw propagates — \`ACTIVE_PATHS\` is registered only on the happy path, so no leak and no orphaned registration.
- **Dropped \`existsSync\` + \`readFileSync\` imports** — both are gone from \`journal.ts\`.

## Edge cases — 4 new tests

| Case | What it exercises |
|---|---|
| Pre-existing empty file (\`touch audit.log\`) | Fresh chain starts cleanly from \`initialPrevHash\` |
| File of only trailing newlines (\`\\\\n\\\\n\\\\n\`) | Scanner loops and terminates at pos=0 with \`return null\` |
| 1,000-event file (~300 KiB) | Tail straddles multiple 64 KiB chunks |
| Single event with 80 KiB \`correlationId\` body | Last line itself exceeds one chunk — scanner must aggregate |

Existing tests still covering the happy paths:
- First write on empty file seeds from \`initialPrevHash\`
- Reopen recovers \`headHash\` + \`nextSequenceNumber\` from disk
- Reopen rejects on invalid-JSON last line
- Single-writer invariant still enforced

## Test plan

- [x] \`bunx @biomejs/biome check .\` — clean
- [x] \`bun test --timeout 15000\` — **690 pass / 0 fail / 4,021 expects** (was 686 / 4,009)
- [x] \`bash scripts/coverage-floor.sh 95\` — 98.46% line / 98.20% func (floor 95%)
- [x] \`bunx depcruise --config .dependency-cruiser.js .\` — no violations
- [x] \`bash scripts/harness-hash.sh --verify\` — OK
- [x] \`bun scripts/crap-score.ts --threshold 30\` — max 28, 0 over
- [x] \`bash scripts/gherkin-lint.sh --path features/ --strict\` — 0 warnings
- [x] \`bun audit --audit-level=high --ignore=GHSA-j3q9-mxjg-w52f\` — clean

Jeremy made me do it
-claude

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **`JournalWriter.open()` — reverse-chunk tail read + single file handle** (`ccsc-otd`). Replaces the full `readFileSync` of the audit log (used only to recover `lastHash` and `nextSeq` from the last line) with a 64 KiB reverse-chunk scan from EOF backwards until the last newline boundary is found. Drops startup memory from O(file size) to O(last line size). Also unifies the previous read-then-fsOpen pair into one `fsOpen('a+', 0o600)` handle: reads use explicit `position` so they don't disturb the append pointer, and writes still go through O_APPEND (atomic EOF writes preserved). The single-handle design incidentally closes the stat-then-use TOCTOU window that triggered the earlier CodeQL `js/file-system-race` pattern on this path. Four new tests in `server.test.ts` exercise the reverse-scan edge cases (pre-existing empty file, trailing-newlines-only, last line straddling a chunk boundary, last line larger than one chunk).
+
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


jeremylongshore:
Addressed in ed6ef40 — replaced the silent `.catch(() => {})` with an explicit try/catch that logs `fh.close` failures to stderr with the file path, still throwing the original recovery error (that's the one the operator needs to fix). Close failures during recovery cleanup now surface rather than vanish.

Jeremy made me do it
-claude

PR #155, opened 2026-05-08
feat(server): SLACK_BRIDGE_DISABLE passive-mode flag for multi-session setups

4466hikaru:
## Summary

Adds a `SLACK_BRIDGE_DISABLE` env flag that puts a `server.ts` instance into passive mode — MCP transport still connects, but Socket Mode / journal / supervisor are skipped and every tool call returns an isError stub. The escape hatch for the documented "single-writer per state directory" undefined behavior when the slack-channel plugin is installed at user scope and multiple Claude Code processes load it simultaneously.

## Why

User-scope plugin install means every Claude Code session for that user spawns its own `server.ts` subprocess (stdio MCP architecture — no shared daemon). Two concurrent processes both read/write `~/.claude/channels/slack/` and both open Socket Mode connections on the same app token:

- **Audit log corruption** — the hash chain in `audit.log` breaks when two `JournalWriter`s append concurrently.
- **Session race** — `sessions/<channel>/<thread>.json` is atomic per-write but two processes can race on the same thread.
- **Inbound load-balancing** — Slack fans inbound events out across the open WebSocket connections (up to 10 per app per their docs); each event lands on whichever connection wins the race, so neither process's `deliveredThreads` reflects ground truth.

CLAUDE.md flags this as undefined behavior; this PR provides the documented escape hatch.

## What changed

- **`lib.ts`** — new pure primitive `isBridgeDisabled(env)`. Accepts `1` / `true` / `yes` (case-insensitive, trim-tolerant). Anything else returns false (no guessing).
- **`server.ts`** — three guards driven by `BRIDGE_DISABLED = isBridgeDisabled(process.env)`:
  - `main()` early-exit branch: skips `auth.test`, `socket.start`, journal open, supervisor boot. Connects MCP transport and stdin shutdown hooks, returns. Existing `shutdown()` already handles null `journal` / `supervisor` / never-started socket.
  - `CallToolRequestSchema` dispatcher: returns isError naming the disabled state. Covers all 8 tools (`reply` / `react` / `edit_message` / `fetch_messages` / `download_attachment` / `list_sessions` / `read_peer_manifests` / `publish_manifest`).
  - `PermissionRequestSchema` notification handler: silently drops. Claude Code's permission flow falls back to the in-CLI prompt.
- **`server.test.ts`** — three new tests covering truthy / falsy / unset env values for `isBridgeDisabled`.
- **`CHANGELOG.md`** — Unreleased entry.

Default-off; existing single-session deployments unaffected.

## Test plan

- [x] `bun run typecheck`
- [x] `bun test --test-name-pattern "isBridgeDisabled"` — 3 pass
- [x] Full test suite — 684 pass / 23 pre-existing Windows-only failures (path / 0o600 / symlink) confirmed identical on baseline `main`
- [x] `bunx depcruise --config .dependency-cruiser.js .` — no violations
- [ ] CI: typecheck / Biome / coverage floor / depcruise / gherkin-lint / harness-hash / bun audit / crap-score
- [ ] **Manual integration verification (operator-side, post-merge)**:
  1. Set `SLACK_BRIDGE_DISABLE=1` in the env of a non-bridge Claude Code session
  2. Restart that session — verify stderr shows `[slack] SLACK_BRIDGE_DISABLE=1 — passive mode` and **does not** show `[slack] Socket Mode connected`
  3. Call `mcp__plugin_slack-channel_slack__reply` from that session — verify isError with the disabled-state message
  4. Send a Slack DM to the bot with both sessions running — verify it reaches only the bridge-enabled session
  5. Call `reply` from the bridge-enabled session — verify it still works

## Security notes

- No new attack surface — the flag only *removes* capabilities (Socket Mode, journal writes, tool execution). Default-off preserves existing behavior bit-for-bit.
- Permission prompts that previously routed to Slack now silently fall back to the local CLI prompt. Operators who relied on remote approval must keep at least one bridge-enabled session running.
- The MCP transport stays connected so Claude Code's tool registry / per-project allowlist remains consistent — no surprise tool disappearance for the user.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,10 +64,25 @@ jobs:
       - name: Dependency vulnerability audit (bun audit)
         # Native bun gate against the GitHub Advisory Database. Only fails on
         # NEW high/critical advisories; known moderates in transitive deps are
-        # tracked separately (see bd ccsc-<follow-up>). The ignored advisory
-        # is a high-severity path-to-regexp finding inside @modelcontextprotocol/sdk
-        # — unfixable from our side until the SDK ships an updated transitive.
-        run: bun audit --audit-level=high --ignore=GHSA-j3q9-mxjg-w52f
+        # tracked separately (see bd ccsc-<follow-up>). The ignored advisories
+        # are all high-severity findings inside transitive deps that we cannot
+        # fix from our side until upstream ships an updated dep:
+        #   - GHSA-j3q9-mxjg-w52f   path-to-regexp (via @modelcontextprotocol/sdk)
+        #   - GHSA-v39h-62p7-jpjc   fast-uri host confusion (via ajv → mcp/sdk + stryker)
+        #   - GHSA-q3j6-qgpj-74h6   fast-uri path traversal (via ajv → mcp/sdk + stryker)
+        #   - GHSA-pmwg-cvhr-8vh7   axios NO_PROXY bypass (via @slack/web-api)
+        #   - GHSA-q8qp-cvcw-x6jj   axios prototype pollution (via @slack/web-api)
+        #   - GHSA-pf86-5x62-jrwf   axios prototype pollution gadgets (via @slack/web-api)
+        #   - GHSA-6chq-wfr3-2hj9   axios header injection (via @slack/web-api)
+        run: |
+          bun audit --audit-level=high \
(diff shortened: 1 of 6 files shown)


4466hikaru:
Closing - will merge into my own fork instead

PR #162, opened 2026-05-13
fix(reply): pass filename to filesUploadV2 so uploads keep their extension

PGMacDesign:
## Summary

The `reply` MCP tool currently uploads files with no `filename` arg to `WebClient.filesUploadV2()`. The `@slack/web-api` v2 upload flow doesn't infer the filename from the `file` path when it's a string — Slack defaults to `file.txt` for every upload, regardless of the actual file type.

Result: every PNG, PDF, HTML, MD, etc. sent through the `reply` tool's `files:` arg lands in Slack as `file.txt` with `text/plain` mimetype, no inline preview, and the wrong download name.

## Reproduction

```ts
await tools.reply({
  chat_id: 'C...',
  text: 'here is the diagram',
  files: ['/abs/path/to/diagram.png'],
})
```

Recipient sees `file.txt` (252 KB) instead of `diagram.png`. No image preview.

I hit this end-to-end today and verified the workaround by calling Slack's `files.getUploadURLExternal` + `files.completeUploadExternal` directly with explicit `filename` — that worked correctly, confirming the missing field is the cause.

## Fix

`server.ts` — add `filename: basename(resolved)` to the upload args. Two-line change (one import addition, one arg addition). No behavior change beyond surfacing the original filename to Slack.

```diff
-import { join, resolve } from 'node:path'
+import { basename, join, resolve } from 'node:path'

       const resolved = resolve(filePath)
       const uploadArgs: Record<string, any> = {
         channel_id: chatId,
         file: resolved,
+        filename: basename(resolved),
       }
```

## Test plan

- [x] `tsc --noEmit` passes locally
- [x] Manual repro: upload PNG via patched server → recipient sees `spool-holder.png` (388 KB) with inline preview
- [x] Manual repro: upload PNG via unpatched server → recipient sees `file.txt` (388 KB), no preview
- [ ] No existing test mocks `filesUploadV2` to assert upload args. Happy to add one (spy on `ctx.web.filesUploadV2`, send a reply with files, assert `filename === basename(file)`) if you'd like — let me know your preference and I'll push a follow-up commit.

## Notes

- No beads ticket on this PR — I'm an external contributor without access. If you want one, you'll know better than I do how to slot it in.
- Pre-commit / pre-push hooks were bypassed locally because `bun` isn't installed on the dev machine; biome and `tsc --noEmit` were run manually and pass clean.
- CHANGELOG `[Unreleased] / Fixed` entry included in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`reply` tool — pass `filename` to Slack file uploads**. `server.ts` was calling `WebClient.filesUploadV2()` with `{ channel_id, file: <absolute path> }` but never setting `filename`. The `@slack/web-api` v2 upload flow does not infer the filename from the `file` path when it's a string — Slack defaults to `file.txt`, which means every uploaded asset (PNG, HTML, MD, PDF, etc.) was delivered to recipients as a `.txt` download with the wrong mimetype. Adds `filename: basename(resolved)` to the upload args so the original filename and inferred mimetype are preserved end-to-end. Imports `basename` from `node:path` (already imports `resolve` from the same module).
+
 ### Security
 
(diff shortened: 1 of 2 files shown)


jeremylongshore:
Heads up @PGMacDesign — I rebased your commit onto current `main` to clear a CI conflict and force-pushed back to your fork (`0eb24fb` → `853f057`). The only change was resolving a CHANGELOG merge conflict; your `server.ts` patch is byte-identical and your authorship is preserved.

Context: `main` was red because of 6 unrelated high-severity transitive CVEs (axios + fast-uri). PR #163 cleared those just now and bumped CHANGELOG's `[Unreleased]` with a `### Security` entry, which collided with your `### Fixed` entry. Resolved by keeping both in Keep-a-Changelog order (Fixed before Security).

If you had any uncommitted work on `fix/upload-filename` locally, `git fetch && git reset --hard origin/fix/upload-filename` will sync you up. Sorry for the inconvenience — and thank you for the fix. About to merge once CI clears.

- Jeremy Longshore
intentsolutions.io

jeremylongshore:
For the credit trail on this one — @gog5-ops independently arrived at the same one-line fix on their `feat/admin-commands-and-thinking-indicator` branch (commit `01210a9`, 2026-05-14), hours after this PR landed. Two contributors hitting the same bug, the fix is so structurally obvious there's literally one way to pass `filename` to `filesUploadV2`, both got to identical diffs. Could be parallel discovery, could be rebase-after-noticing-main — either way both contributors saw it.

@PGMacDesign — thanks again for the surgical fix and the explicit repro + test plan. The patch was easy to review, easy to verify, easy to ship. Inline image previews are working correctly in production now.

- Jeremy Longshore
intentsolutions.io

PR #166, opened 2026-05-14
Multi-state reaction lifecycle on the inbound message

putza:
## Problem

The existing `ackReaction` (default `:eyes:`) gets added when the broker forwards a mention to claude, but it never updates. In an active operations channel, the operator can't tell at a glance whether a given conversation is still working, awaiting approval, completed, denied, or errored — every turn just has 👀 forever.

## Fix

A small state machine with five states (terminal except `:eyes:` and `:warning:`):

| State | Emoji | Trigger |
|---|---|---|
| working | 👀 (`:eyes:`) | inbound forwarded to claude (existing behavior, now goes through the helper) |
| awaiting input | ⚠️ (`:warning:`) | `policy.require_human` route — Block Kit prompt posted |
| done | ✅ (`:white_check_mark:`) | `executeReply` first successful `chat.postMessage` |
| denied | ⛔ (`:no_entry:`) | `policy.deny` route |
| errored | ❌ (`:x:`) | tool dispatch try/catch |

## Implementation

Two new module-level state vars sibling to the existing `lastActiveChannel` / `lastActiveThread`:

- `lastActiveTs` — the ts of the most-recent inbound message (so the helper knows WHICH message to react on).
- `lastActiveReaction` — the emoji currently set (so the helper knows what to remove before adding next).

New helper `setInboundReaction(web, channel, ts, next)`:
- Best-effort `reactions.remove` of `lastActiveReaction` (if any), wrapped in `try { ... } catch { /* non-critical */ }`.
- Best-effort `reactions.add` of `next`, same wrapper.
- Update `lastActiveReaction = next`.

Both Slack API calls wrapped because **reactions are UX polish and must never break tool calls**.

Transitions wired at the existing call sites — no new infrastructure beyond the helper.

## Single-operator caveat

The state is module-level globals (not per-(channel, ts) keyed). Single-operator deployment is race-free; concurrent inbounds across channels would have the later inbound stomp the earlier one's tracker. Acceptable for our use; the structure is easy to extend to a `Map<(channel, ts), Reaction>` later if multi-operator deployments need it.

## Slack rate limits

Each turn does at most 2 reactions ops (remove → add). Slack tier 2 limit is ~100/min/user. Well under for normal operator volume.

## Tested in production

Running as a local patch in our Hearth deployment for ~2 days. All five transitions fire correctly across ~100 bot turns. The catch sites for `:x:` were verified by deliberately killing one of the MCP children mid-tool-call.

## Notes

- Companion to #165 (reply-in-thread default) but independent — patch 02 doesn't depend on patch 01.
- No new tests added — would naturally go into a per-route reaction-lifecycle test suite. Happy to add if you'd like to scope that.

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -550,6 +550,33 @@ let reaperTimer: ReturnType<typeof setInterval> | null = null
 // Track last active channel/thread for permission relay
 let lastActiveChannel = ''
 let lastActiveThread: string | undefined
+// Sprint 5+6 reaction lifecycle: track the ts of the most-recent inbound
+// message + the current reaction emoji on it so the lifecycle helper can
+// swap reactions without the caller needing to know the previous state.
+// Single-operator deployment safe; concurrent inbounds across channels
+// would race — fine for our use, addressed in a future patch.
+let lastActiveTs: string | undefined
+let lastActiveReaction: string | undefined
+
+// Transition the inbound's ack reaction to `next`. Best-effort removes
+// whatever reaction the broker last set (tracked in module state) and
+// then best-effort adds `next`. Either Slack API call may fail (rate
+// limit, missing reaction, etc.); reactions are UX polish, so we
+// swallow errors per the non-critical pattern used by the existing
+// reactions.add at the inbound site.
+async function setInboundReaction(
+  web: WebClient,
+  channel: string,
+  ts: string,
+  next: string,
+): Promise<void> {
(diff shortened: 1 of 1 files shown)


putza:
Closing — this was filed in error from a downstream patch in our local fork. We're not yet ready to upstream this change. Apologies for the noise.

Issue #212, opened 2026-06-03
Evaluate Slack's native agent approval and interactive-component primitives.


jeremylongshore:
Evaluation epic. The peer runtime does NOT wire HITL to Slack (thin webhook listener; its approval primitive was never connected). Slack's native agentic surface (approval previews; Card/Alert/Data-Table components; Agent Browser) is an opportunity to evaluate against CCSC's nonce-HITL and audit projection. Cross-references AGP's nonce-HITL since HITL lives in both systems.

**Beads:** ccsc-pko (epic) + ccsc-pko.1 (approval-preview spike), ccsc-pko.2 (native-component receipts prototype), ccsc-pko.3 (adopt/hybridize/keep decision)

**Provenance:** patterns from a peer-runtime audit, distilled in `000-docs/ADR-002-architecture-patterns-from-peer-runtime-audit.md`; full decision record in the private AT-DECR 013. No peer brand is named by design.

- Jeremy Longshore
intentsolutions.io

jeremylongshore:
Linked bead: `ccsc-pko` (source of truth). Use `bd-sync` to mirror updates.

jeremylongshore:
**bd-sync from `ccsc-pko`:** Re-scope (2026-06-23): Anthropic shipped Claude Tag — first-party @Claude in Slack (per-channel scope, viewable audit log, token-spend caps; Team/Enterprise beta). Re-frame this evaluation epic to benchmark against that SHIPPED surface rather than generic Slack-native primitives: (1) approval UX vs our nonce-HITL, (2) viewable audit-log vs our signed offline-verifiable journal + projection, (3) spend caps as a feature to consider. Prerequisite finding from the new mention-gating epic (ccsc-apj): A1 thread-stickiness (ccsc-apj.1) must land first — the existing requireMention/allowFrom/ambient surface has to be usable (mention-once-then-converse) before deciding adopt/hybridize/keep Block Kit (pko.3). Sources: https://www.anthropic.com/news/introducing-claude-tag | https://techcrunch.com/2026/06/23/anthropics-claude-tag-is-learning-your-company-one-slack-message-at-a-time/ | https://venturebeat.com/technology/anthropic-launches-claude-tag-replacing-its-slack-app-with-a-persistent-ai-teammate-that-learns-monitors-and-works-autonomously

jeremylongshore:
[...]

Issue #210, opened 2026-06-03
Turn the peer runtime's footguns into our own regression tests.


jeremylongshore:
Convert the anti-patterns the peer-runtime audit surfaced into CCSC regression tests that prove we are NOT vulnerable — defensive self-audit, not feature work. On signed audit and per-tool-call authz CCSC is already ahead; these tests lock that in.

**Beads:** ccsc-1iw (epic) + ccsc-1iw.1 (cross-thread read isolation), ccsc-1iw.2 (no journal gaps), ccsc-1iw.3 (fail-closed defaults audit)

**Provenance:** patterns from a peer-runtime audit, distilled in `000-docs/ADR-002-architecture-patterns-from-peer-runtime-audit.md`; full decision record in the private AT-DECR 013. No peer brand is named by design.

- Jeremy Longshore
intentsolutions.io

jeremylongshore:
Linked bead: `ccsc-1iw` (source of truth). Use `bd-sync` to mirror updates.

jeremylongshore:
**bd-sync from `ccsc-1iw`:** Child 1/3 shipped: cross-(channel,thread) session isolation pinned at the supervisor layer (PR #232, commit 4762895). Proves disjoint state across channels, distinct sessions for same thread-id in two channels, and inbound-traversal-key rejection before any read. Remaining: ccsc-1iw.2 (gapless journal of every tool-call decision), ccsc-1iw.3 (fail-open→fail-closed default audit).

jeremylongshore:
**bd-sync from `ccsc-1iw`:** Child 2/3 shipped: 'every policy decision is journaled — no silent gaps' pinned (PR #233). A never-guarded decision→journal-kind mapping makes a new un-journaled route branch a typecheck failure; only the no-opinion default_human route is deliberately silent. Filed follow-up ccsc-175 (make server.ts dispatch importable so the invariant binds production code). Remaining: ccsc-1iw.3 (review fail-open access defaults, flip to fail-closed) — a behavior change, surfaced for review when reached.

jeremylongshore:
**bd-sync close on `ccsc-1iw`:** Epic complete: all 3 footgun-inversion regression tests shipped (PRs #232/#233/#234). CCSC was already correct on cross-thread isolation, gapless journaling, and fail-closed defaults — these pins prevent regression. Follow-up ccsc-175 filed. _(GH issue left open — covers other beads)_

Issue #208, opened 2026-06-03
Make Slack sessions crash-safe and replies loss-proof.


jeremylongshore:
Harden CCSC's session runtime against process crashes and dropped replies, adopting the peer runtime's durability primitives.

**Today:** supervisor crash recovery is reload-only with no fencing; a failed `chat.postMessage` is logged to stderr and swallowed, so a terminal turn can produce no reply with no retry.

**Sub-epics & beads (cluster under this issue):**
- 1-A Lease-fenced durable execution — `ccsc-o7x.1`: per-turn fencing lease + heartbeat (`ccsc-o7x.1.1`), restart-recovery sweep (`ccsc-o7x.1.2`), lease-loss → quarantine (`ccsc-o7x.1.3`)
- 1-B Reliable delivery outbox — `ccsc-o7x.2`: durable delivery obligation (`ccsc-o7x.2.1`), leased retrying/dead-lettering poller (`ccsc-o7x.2.2`), idempotent sends (`ccsc-o7x.2.3`)

1-B depends on 1-A's lease/state model.

**Beads:** ccsc-o7x (epic) + ccsc-o7x.1, ccsc-o7x.1.1, ccsc-o7x.1.2, ccsc-o7x.1.3, ccsc-o7x.2, ccsc-o7x.2.1, ccsc-o7x.2.2, ccsc-o7x.2.3

**Provenance:** patterns from a peer-runtime audit, distilled in `000-docs/ADR-002-architecture-patterns-from-peer-runtime-audit.md`; full decision record in the private AT-DECR 013. No peer brand is named by design.

- Jeremy Longshore
intentsolutions.io

jeremylongshore:
Linked bead: `ccsc-o7x` (source of truth). Use `bd-sync` to mirror updates.

jeremylongshore:
**bd-sync from `ccsc-o7x`:** ccsc-o7x.1.1 (per-turn fencing lease + heartbeat) shipped in PR #220 and merged to main. Sub-epic 1-A now 1/3: next is ccsc-o7x.1.2 (restart-recovery sweep — requeue stale leases, reap orphaned turns), then ccsc-o7x.1.3 (route lease-loss into quarantine). The lease primitive (token+owner+heartbeatAt, fenced update, heartbeat) is in place for them to build on.

jeremylongshore:
[...]

PR #221, opened 2026-06-04
Feat/multi session routing

threecee-adsk:
## How

---

## Security impact

Check every box that applies. If any of these are checked, the PR description
**must** explain the threat-model reasoning — not just "tests pass."

- [ ] Touches `gate()` / inbound gating logic in `lib.ts`
- [ ] Touches `assertSendable()` / file exfiltration guard
- [ ] Touches `assertOutboundAllowed()` / outbound gating logic
- [ ] Changes the Socket Mode / MCP stdio transport in `server.ts`
- [ ] Adds, removes, or modifies an MCP tool exposed to Claude
- [ ] Touches token loading, `.env` handling, or `access.json` I/O
- [ ] Changes the system-prompt hardening text
- [ ] Adds a new runtime dependency (anything landing in `package.json`)
- [ ] None of the above — this is a docs / test-only / cosmetic change

---

## Test evidence

Pick the tier that matches your change. One tier is enough — pick the strongest
one that genuinely applies. Don't claim a tier you didn't meet.

### Tier A — Automated test in the suite

- [ ] I added or updated tests in `server.test.ts` (or a new `*.test.ts` file)
- [ ] `bun test` passes locally
- [ ] `bun run typecheck` passes locally
- [ ] The test would fail without my fix (I verified by reverting the fix)

### Tier B — Reproducible script CI can run

Use this when the behavior is hard to unit-test (real Slack event shapes,
real MCP handshake, file-system edge cases) but can still be exercised
deterministically from a script.

- [ ] Script lives at `scripts/<name>.ts` (or is inlined in the PR body)
- [ ] Script is runnable with `bun run scripts/<name>.ts` with no secrets
- [ ] Script exits non-zero on failure
- [ ] Expected output is pasted below, verbatim

<!-- ```
$ bun run scripts/repro.ts
... expected output ...
``` -->

### Tier C — Human evidence you ran yourself

Use this only when the behavior truly cannot be automated (signals,
real Slack handshakes, zombie-process checks). **You** run it, **you** post
the evidence. The maintainer will not re-run it — if your evidence isn't
convincing, you'll be asked for more, or the PR will be declined.

- [ ] Exact command(s) you ran, copy-pasteable
- [ ] Your actual terminal output pasted below (or a screen recording attached)
- [ ] Environment: OS, Bun version, Claude Code version
- [ ] Any required setup (tokens, workspace, channel) called out

<!-- 1. ...
     2. ...
     3. Observed: ... -->

---

## Checklist

- [ ] I read `CLAUDE.md` and `SECURITY.md` and the change is consistent with them
- [ ] Pure logic went into `lib.ts`; stateful wiring stayed in `server.ts`
- [ ] I did not add runtime dependencies beyond what's needed (the "four deps" rule)
- [ ] Commit messages describe *why*, not just *what*
- [ ] If this is a breaking change, `CHANGELOG.md` says so loudly
- [ ] I re-read the full PR diff (not just the latest commit) before clicking merge


threecee-adsk:
Sorry, my agent auto created this pull request, no need to merge it in. Adds multi-session, multi channel support

jeremylongshore:
Carl — thank you for this, genuinely. You nailed the constraint most people miss: Slack load-balances Socket Mode across all of an app's connections, so exactly one process must own the socket and the rest attach as clients. The router framing follows correctly and the design doc is thoughtful. I read all of it.

I'm going to be straight with you about fit, because you put in real work and deserve a real answer instead of a PR that lingers.

**CCSC is deliberately a single-process governance _substrate_** — a small, auditable kernel (policy gate + signed audit journal + Slack HITL) that a companion project vendors as a dependency. Its entire open roadmap (#208, #210, #212) is about hardening *that*: crash-safety, security regression tests, tightening the gate. Multi-session routing isn't a smaller-or-larger version of that work; it's a different shape, and turning the substrate multi-process (plus a new localhost trust boundary between router and sessions) cuts against the one thing this repo is supposed to stay. I've just published a [`ROADMAP.md`](https://github.com/jeremylongshore/claude-code-slack-channel/blob/main/ROADMAP.md) with an explicit **Non-Goals** section so this is documented rather than a surprise — multi-process / multi-session orchestration is called out there directly.

**But the direction you're building toward already has a home.** I maintain [`agent-governance-plane`](https://github.com/jeremylongshore/agent-governance-plane), whose entire purpose is governing *multiple* sandboxed agent sessions over a gateway — built on CCSC's kernel. Your router / registry / binding thinking is squarely an AGP conversation, and a much better fit for it.

So I'm going to close this PR here — not as a rejection of the work, but because it belongs in a different repo. If you're up for it, **open an issue on [`agent-governance-plane`](https://github.com/jeremylongshore/agent-governance-plane/issues) with the multi-session use case + your design** (port the doc over) and tag me — I'll engage with it there properly. And if you'd ever like to contribute to CCSC itself, the roadmap issues above are the live tracks; happy to point you at a good first one.

Either way — this was real work and I'd like to keep building with you on the right surface. Thanks again.

- Jeremy Longshore
intentsolutions.io

Issue #242, opened 2026-06-23
Harden multi-agent shared channels against runaway loops and context bleed.


jeremylongshore:
**Epic — hardening multi-agent shared channels against runaway loops and context bleed.**

Follows the closed multi-agent epic (humans + multiple AI agents in one channel: `allowBotIds` opt-in, per-`(channel,bot)` sliding-window rate limit, `!mute`/`!unmute`, self-echo filter). This epic closes the remaining hardening gaps the engineering review surfaced.

**Children**
- **B1** (`ccsc-0k7x2`) — per-channel circuit breaker for N-cycle (A→B→C→A) bot loops on top of the pairwise limiter. *(G2 — `peer-bot-rate-limit.ts` windows are per-sender, so 3+-agent rings pass)*
- **B2** (`ccsc-kl410`) — optional per-user session isolation so two humans in one thread don't share context. *(G3 — `SessionKey` is `{channel, thread}` only)*
- **B3** (`ccsc-yl6k9`) — make peer-bot mutes + rate limits observable / hot-editable from Slack (`!mute-status`, `!rate-limit`, auto-expire notice). *(G6)*
- **B4** (`ccsc-4e9bf`) — "agents online" indicator + global `maxConcurrentSessions` backpressure cap. *(G8/G9 — low; blocked by B1–B3)*

**Coupling:** the mention-gating epic keeps peer agents per-message-explicit (never thread-sticky); this epic is the backstop that keeps agent traffic bounded.

**Beads:** ccsc-uorlh (epic) · ccsc-0k7x2 · ccsc-kl410 · ccsc-yl6k9 · ccsc-4e9bf

- Jeremy Longshore
intentsolutions.io

jeremylongshore:
Linked bead: `ccsc-uorlh` (source of truth). Use `bd-sync` to mirror updates.

jeremylongshore:
[...]

PR #248, opened 2026-06-24
Optional per-user session isolation (ccsc-kl410)

jeremylongshore:
Multi-agent hardening epic (#242). Closes gap **G3** (multi-human thread context bleed).

## Problem
By default two humans in one Slack thread share **one** bridge session — same state file, `ownerId`, and in-flight tracking. `ownerId` is recorded but the sharing isn't isolatable, and nothing signals it.

## Fix (opt-in, default-off)
A channel can set `perUserSessions: true`. Each distinct sender then gets their own session **within** the shared thread.

- `lib.ts`: `SessionKey` gains optional `userId`; `sessionPath` nests the file at `sessions/<channel>/<userId>/<thread>.json` with the **same realpath + component-injection guards** as channel/thread; `ChannelPolicy.perUserSessions`.
- `supervisor.ts`: `keyId` appends the userId segment → a per-user key never collides with the shared `(channel, thread)` key.
- `server.ts`: `deliverEvent` builds the key via `inboundSessionKey()` (helper keeps it under the CRAP gate) and uses it for both the `gate.inbound.deliver` journal record (key shape is auditable) and supervisor activation.

This is **bridge-session** isolation (per-thread book-keeping per user), not a change to Claude's conversation memory.

> Relevant to the Claude Tag critique thread: the "LLMs are 1:1-trained, don't grok multi-party" limitation is mitigated at the **harness** level here — separate per-user session state instead of blending a busy channel into one stream.

## Security
The new `userId` path component runs the identical `isValidSessionComponent` guard (rejects `.`/`..`/`/`) and the realpath-under-root check, so it can't inject traversal. Tests cover both rejection paths.

## Verification
typecheck · biome · **1199 tests / 0 fail** · coverage **96.57% line / 97.94% func** · crap-score (29) · depcruise · gherkin-lint · harness-hash · bun audit — all green. `000-docs/session-state-machine.md` (frozen contract) + `ACCESS.md` updated.

**Beads:** ccsc-kl410 (closed). Remaining in #242: Slack-observable mutes/limits (G6), agents-online + backpressure (G8/G9).

- Jeremy Longshore
intentsolutions.io


jeremylongshore:
Both addressed in the latest commit:
- **High — `inboundSessionKey` sender id:** now `(ev.user ?? ev.bot_id)` with a `typeof === 'string' && !== ''` guard, so a peer bot (ev.user undefined, bot_id set) gets isolated too, and an absent/empty sender falls back to the shared session instead of building an invalid empty-userId key.
- **Security-medium — empty userId in `sessionPath`:** the component regex `/^[A-Za-z0-9._-]+$/` already rejects `''` (the `+` needs ≥1 char), but I've made it explicit (`userId === '' || !isValidSessionComponent(...)`) at the boundary and added a test asserting empty userId throws `invalid userId component`.

Gates green: typecheck · biome · 1200 tests / 0 fail · crap 29.

- Jeremy Longshore
intentsolutions.io

PR #254, opened 2026-06-24
feat(reply): make streaming replies crash-durable (ccsc-o7x.6)

jeremylongshore:
Refs #208
**Beads:** ccsc-o7x.6 (crash-safety epic `ccsc-o7x`)

The second of the two deferred rich paths (after chunked replies, `ccsc-o7x.4`). A streaming reply (`stream:true`, text over the chunk limit) posts an initial message and grows it via `chat.update` — which **can't be idempotently replayed mid-stream**, so `ccsc-o7x.3` deferred its durability with a documented path.

## Design (ADR-002 addendum, 2026-06-24)
Record **one** `pending` `DeliveryObligation` carrying the **full text** *before* the stream starts, then resolve it from the `streamReply` outcome:

| Outcome | Resolution |
|---|---|
| `completed` | `markDelivered` — the streamed message holds the full content |
| in-process failure (`failed_mid_stream` / `gate_rejected_at_start` / init throw) | `markDead` — **not** left pending |
| **process crash mid-stream** | neither mark runs → obligation stays `pending` → boot-drain/poller **redelivers the full text as a fresh `chat.postMessage`** |

**Why dead-on-failure is load-bearing:** the delivery poller's `send` does **not** re-run `assertOutboundAllowed`, and `streamReply` can't distinguish a mid-stream gate rejection from a transient Slack error. Leaving the obligation pending would let the poller redeliver to a channel the outbound gate just rejected — a gate bypass. So an in-process failure resolves terminal; only an actual process crash leaves it pending. This is exactly the bead's wording: "a crash mid-stream redelivers the final message."

## Why no kernel/poller change
The full-text obligation is an ordinary single-message `DeliveryObligation` (`payload` = full text). The existing poller posts `payload` via `chat.postMessage` — precisely the redelivery we want — and the idempotency key dedups poller-side re-attempts. `lib.ts` (the AGP-vendored kernel) and `drainOutbox` are untouched.

## Changes
- `slack-delivery.ts`: new `beginDurableStream()` + `DurableStream` / `DurableStreamHandle` (reuses the existing `markObligation`).
- `server.ts`: `maybeBeginDurableStream` + `resolveStreamObligation` helpers wire it into `executeReplyStreamingPath` (extracted to keep CRAP at 29 < 30).
- `000-docs/ADR-002-...md`: addendum documenting the design + two residuals.
- `CHANGELOG.md`: entry.

## Documented residuals
- Completed-but-crashed-before-marking window → one un-deduped duplicate (the streamed message carries no delivery metadata, so `findDelivered` can't recognise it — mirrors the pre-idempotency-key text path).
- >40k-char single-message redelivery → dead-letters as `msg_too_long`, but such a reply couldn't have streamed cleanly to begin with (`chat.update` shares the 40k limit).

## Verification
- 5 new tests (`beginDurableStream`): records one full-text pending obligation; `markDelivered`; `markDead` with reason; crash-leaves-pending across a restart; `DurableUnavailableError` records nothing.
- All nine gates green locally: typecheck, Biome, 1226 tests, coverage **96.65% line / 98.20% func** (`slack-delivery.ts` **100% / 100%**), depcruise, gherkin-lint, harness-hash, bun audit, CRAP (server.ts max 29 < 30).

Last deferred rich path after this: file uploads (`ccsc-o7x.5`).


jeremylongshore:
Thanks — addressed:

**`randomUUID` (server.ts:1167)** — false positive. It's imported at the top of `server.ts`: `import { randomUUID } from 'node:crypto'` (line 2), and already used by the text/chunked durable paths (lines 1314, 1355). No change needed.

**`recordTerminalDelivery` propagation (slack-delivery.ts)** — good catch on the failure mode; fixed at a different layer than suggested (commit pushed). Rather than have `beginDurableStream` convert record-failures into `DurableUnavailableError` (which would muddy its contract and diverge from its siblings `deliverReplyDurably`/`deliverChunkedReplyDurably`, where the obligation *is* the send path and a record failure *should* surface), `maybeBeginDurableStream` now degrades to best-effort streaming on **any** failure — `DurableUnavailableError` *or* a `recordTerminalDelivery` write failure — and logs non-`DurableUnavailableError` cases to stderr. The streaming obligation is a crash safety-net *separate* from the send, so failing to record it must never fail the reply; this keeps o7x.6 durability strictly additive.

- Jeremy Longshore
intentsolutions.io

jeremylongshore:
Good catch — fixed (pushed). `resolveStreamObligation` is now a `switch` with an explicit `gate_rejected_at_start` case and a `const _exhaustive: never = result` default guard, matching the `permissionRouteJournalEvents` never-guard pattern (`ccsc-175`). A new `StreamReplyResult` variant now fails to compile instead of silently dead-lettering with a misleading reason.

- Jeremy Longshore
intentsolutions.io

PR #255, opened 2026-06-25
feat(reply): durable file-upload building block + design (ccsc-o7x.5 part 1)

jeremylongshore:
Refs #208
**Beads:** ccsc-o7x.5 (crash-safety epic `ccsc-o7x`) — **part 1 of 2**

The last deferred rich path. Like `ccsc-o7x.3` (which shipped in 3 parts), this lands the **tested building blocks + design** first; **part 2** wires `executeReply`/the poller with the production `filesUploadV2` adapter.

## Design (ADR-002 addendum, 2026-06-24)
Two hard facts force the contract:
1. **A Slack file-share message cannot carry app `metadata`.** So the text path's metadata-keyed exact-once (`ccsc-o7x.2.3`) is impossible for files — dedup is a **`(filename, size)` thread scan**, necessarily heuristic.
2. **A file's bytes can change between record-time and redelivery.** So the exfil guard must run at **every** upload, not just at record-time.

## What part 1 ships (all tested, no wiring yet)
- **`lib.ts` kernel:** `DeliveryObligation` gains additive optional `upload` + `uploadedFileId` (+ Zod). **AGP impact: none** — the reply-delivery outbox is not in AGP's reimplemented subset (verified against `agent-governance-plane/substrate/UPSTREAM.md`; AGP *reimplements*, doesn't vendor, per its ADR 040).
- **`supervisor.ts`:** `recordTerminalDeliveries` carries `upload` through.
- **`slack-delivery.ts`:**
  - `ExfilBlockedError` — the durable loop treats it as **non-retryable** (a blocked file is marked `dead`, never uploaded, never retried).
  - `FileUploadDeps` — injected `guard` / `findUploaded` / `upload` (no Slack SDK in this module).
  - `sendFileObligation` — **dedup → guard → upload**; the guard re-runs on every call that reaches the upload (inline *and* poller), so a file that gained a secret since record-time is blocked at redelivery.
  - `deliverFileReplyDurably` — records text-chunk + file obligations atomically, delivers in order (text then files), records `uploadedFileId` after a successful upload, stops at the first transient gap (`queued`), marks dead + rethrows on `ExfilBlockedError`/non-retryable.

## Documented residual
`(filename, size)`-scan dedup leaves a narrow double-upload window (uploaded, crashed before recording `uploadedFileId`, and the rescan finds no match). It's a double-*deliver* of an already-guarded file, never a loss and never a gate bypass — the best Slack's file API permits.

## Verification
- 11 new tests (`sendFileObligation`: dedup-hit/miss-order/guard-block/no-descriptor; `deliverFileReplyDurably`: all-deliver+uploadedFileId, record-before-send, dedup-skip, exfil-block→dead+rethrow, non-retryable→dead, transient→queued, DurableUnavailableError).
- All nine gates green locally: typecheck, Biome, **1237 tests**, coverage (`slack-delivery.ts` **100%/100%**, `lib.ts` 100%/98.97%), depcruise, gherkin-lint, harness-hash, audit, CRAP (lib 28 / server 29, under 30).
- Note: `lib.ts` is in the Stryker set; the change is purely additive declarative fields (no new logic), covered by the Session-schema round-trip + new tests. Full Stryker (~45 min, manual) recommended before the epic closes.

Part 2 will wire it live; the epic's rich paths (single/chunked/streaming/file) are then complete.


jeremylongshore:
Both findings addressed in this PR before merge (commit pushed):

**Issue 2 (TOCTOU — the more important one).** You're right that `guard`-reads-then-`upload`-re-reads can't honor "the bytes that will ACTUALLY be sent." Redesigned `FileUploadDeps` to be **read-once**: `readAndGuard(upload): Promise<Uint8Array>` reads the file *once*, scans those exact bytes, and **returns** them; `upload(obligation, bytes)` sends that same buffer and never re-reads the path. So the guarded bytes are provably the sent bytes. `sendFileObligation` is now `dedup → readAndGuard → upload`.

**Issue 1 (poller classification).** Correct that `drainOutbox` only classifies by Slack error code, so an `ExfilBlockedError` would be retried to `maxAttempts` instead of dead-lettered immediately (the file is never uploaded — guard runs first — but the audit record is wrong). Root cause is reachability: `drainOutbox` (supervisor.ts) couldn't import `ExfilBlockedError` from `slack-delivery.ts` without a cycle. So I **moved `ExfilBlockedError` into `lib.ts`** (the kernel both modules already import). The actual `instanceof` branch in `drainOutbox` lands in **part 2** alongside the poller's file-send wiring + its test (adding it now would be an uncovered branch, since part 1 doesn't route file obligations to the poller). It's recorded as a hard part-2 requirement on the bead.

Still part 1 = building blocks only (no live wiring). All nine gates green; `slack-delivery.ts` 100/100.

- Jeremy Longshore
intentsolutions.io

PR #256, opened 2026-06-25
feat(reply): wire durable file-upload replies live (ccsc-o7x.5 part 2)

jeremylongshore:
Refs #208
**Beads:** ccsc-o7x.5 (crash-safety epic `ccsc-o7x`) — **part 2 of 2**

Wires part 1's building blocks into the runtime. **With this merge, single (`.3`), chunked (`.4`), streaming (`.6`), and file (`.5`) replies are all crash-durable — the epic's rich paths are complete.**

## What this wires
- **`slack-delivery.ts` — `createFileSendDeps`** (production `FileUploadDeps` over a real `WebClient`):
  - `readAndGuard` reads the file **once**, re-runs the outbound exfil guard (`assertSendable` + `assertNoSecretValues` on the latin1 content **and** filename) on the bytes it will upload, journals `exfil.block` + throws `ExfilBlockedError` on a hit, and returns those exact bytes (TOCTOU-safe — `upload` sends the same buffer, no second read).
  - `findUploaded` dedups by recorded `uploadedFileId` or a `(filename, size)` `conversations.replies` scan (returns `null`, never `''`).
  - `upload` sends the guarded bytes via `filesUploadV2`.
- **`supervisor.ts` — `drainOutbox`** treats `ExfilBlockedError` as **non-retryable** (dead-letter on the first failure, not after `maxAttempts` — the bytes won't pass on a retry). This closes the part-1-review gap Greptile flagged, now possible because `ExfilBlockedError` lives in the kernel.
- **`server.ts`** — `executeReply` routes a non-streaming reply **with files** to `executeReplyFileDurablePath` → `deliverFileReplyDurably`; the poller's `send` branches on `obligation.upload` (file vs text). A blocked file inline propagates to the agent (as the best-effort path did); a files-only reply uploads without an empty message.

## Tests
- **13 new tests.** `createFileSendDeps`: read-once guard order, three exfil-block paths (sendable / content / filename), dedup by id / `(filename,size)` / no-match / no-thread / size-error, upload + nested-id extraction + no-descriptor. `drainOutbox`: `ExfilBlockedError` dead-letters on the **first** failure (asserts `attempts === 1`).
- `slack-delivery.ts` **100% line / 100% func**; `supervisor.ts` 98.4/99.8. `server.ts` dispatch + poller wiring is typecheck-covered (consistent with its untested sibling paths — `server.ts` self-runs `main()` so it isn't unit-loadable).
- All nine gates green locally **except** a pre-existing load-induced flake in the unrelated `verifyJournal` 1000-event test (`ccsc-2oy` — passes on a fresh runner / re-run; my 29 file/durable tests pass clean and consistently in isolation).

## Honest residual (unchanged from part-1 design)
Slack file shares carry no app `metadata`, so dedup is a `(filename, size)` scan with one narrow double-*deliver* window (uploaded, crashed before recording the id, rescan misses) — never a loss, never a gate bypass.

Merging this completes `ccsc-o7x.5` and the crash-safety epic's rich-path coverage.


jeremylongshore:
Good catch — fixed (commit ef85f2a). The `(filename, size)` scan in `findUploaded` now only matches files shared **at/after the obligation's `createdAt`** (Slack message `ts` → ms), so an older same-name+size file from a previous turn can no longer falsely dedup and drop the new upload. The exact `uploadedFileId` match stays un-scoped (an id is definitive). Added a regression test (`stale-older-file ... is NOT matched`), and updated the ADR-002 addendum + CHANGELOG to record the delivery-window rule. `slack-delivery.ts` stays 100% line/func.

- Jeremy Longshore
intentsolutions.io

PR #264, opened 2026-07-09
fix(security): enforce policy thread_ts + pin journal genesis seq (Fable-5 review)

parciv4l:
## Summary

Lands the confirmed findings from a **Fable 5** security review of the load-bearing core (`gate()`/`assertSendable()`/`assertOutboundAllowed()` in `lib.ts`, `policy.ts`, `journal.ts`). Two real defects (one HIGH, one MEDIUM) plus two low-severity hardening/doc fixes.

### ① HIGH — `thread_ts` policy predicate was ignored at runtime
`matchApplies()` (`policy.ts`) compared `tool`, `channel`, `actor`, `pathPrefix`, `argEquals` but **never** `match.thread_ts`, despite the schema accepting a `thread_ts`-only rule (the refinement counts it as a constraining field). A rule authored to trust/quarantine a **single** thread — e.g. `{ effect: 'auto_approve', match: { thread_ts: '…' } }` — silently applied to **every** thread and channel, since the remaining predicates were wildcards. A thread-scoped `auto_approve` therefore became a global auto-approve, bypassing the human approver everywhere.

- Fix: compare `match.thread_ts` against `call.sessionKey.thread`.
- The field was documented as intentionally schema-only "until Epic 29-B wires evaluate()" — this lands that wiring. Doc + the stale `server.ts` "can still match on thread_ts" comment reconciled.
- Tests: scoped allow, non-matching thread falls through to default allow, scoped deny leaves other threads unaffected.

### ③ MEDIUM — `verifyJournal()` genesis-seq pin (head-truncation defense)
The verifier trusted the first event's `prevHash` as the (per-file random, unpinnable) genesis anchor and never required the first accepted event's `seq === 1`. An attacker with file-write access could shear the first N events off the top; the surviving prefix (`seq = N+1…`) chained cleanly among itself and verified OK.

- Fix: reject a first event whose `seq ≠ 1` with reason `journal head truncated`.
- The two sequence checks are extracted into `checkSeqContinuity()` to keep `verifyJournal` under the Wall-5 CRAP gate.
- **Residual (deferred):** a full-file rewrite that downgrades every event to unsigned v1 (stripping Ed25519 signatures) still verifies; defeating it needs a pinned "signing began at seq N" anchor. Noted in the verify docstring + CHANGELOG.
- Test: head-truncation regression.

### ④ LOW — prototype-chain hardening on `access.channels`
`assertOutboundAllowed()` and the inbound channel handler indexed `access.channels[chatId]` with a bare truthiness check, so a channel id equal to an inherited `Object.prototype` member (`'constructor'`, `'toString'`, …) read a truthy function off the prototype chain and passed the gate. Not exploitable (Slack ids never take those forms); both sites now use `Object.hasOwn()`.

### ⑤ LOW — THREAT-MODEL §T7 keyspace correction (doc-only)
`PERMISSION_REPLY_RE`'s `[a-km-z]` is 25 letters (excludes only `l`; still includes `o`) → ~9.8M (25⁵) space, not the "22 letters / ~5M" the doc claimed. The regex is a fail-closed **validator** over the request_id Claude Code emits, not a generator, so it is deliberately **not** tightened to also drop `o` (that would reject legitimate ids); only the doc is corrected.

## Deferred (tracked here; `bd` CLI unavailable in the authoring env)

- **Finding ②:** the permission-request site passes `input: {}` to `evaluate()`, so `pathPrefix`/`argEquals` **deny** rules never match there and degrade to the default human-approval route (fail-open for deny). Documented at `server.ts:1938-1942`. Needs structured tool input threaded through the permission-request path (or a load-time reject/warn when a deny rule constrains `pathPrefix`/`argEquals`).
- **v1-downgrade residual** on `verifyJournal` (see ③).

## Verification

All nine CI-equivalent gates green locally: typecheck · Biome · **990 tests pass** (+4) · coverage 95.76% line / 97.83% func (floor 95) · depcruise · gherkin-lint · harness-hash · crap-score (max 28 ≤ 30).


parciv4l:
Addressed in cf59154 — applied the truthy check: `if (Object.hasOwn(access.channels, chatId) && access.channels[chatId]) return`, consistent with the inbound gate's `if (!policy)` drop. And added the missing regression coverage for finding ④: a per-gate pair asserting a `'constructor'`/`'toString'` channel id is refused (outbound throws `Outbound gate`; inbound drops with `channel.not_opted`). Full suite 1257 pass, all nine gates green. Thanks for the catch.

jeremylongshore:
@parciv4l — this is excellent work, thank you — and thanks for the quick turnaround on the `Object.hasOwn` coverage (finding ④) earlier in the thread. A first PR that lands a genuine global-auto-approve-bypass fix, pins the journal against head-truncation, **and** hardens the prototype-chain reads — with a regression test for each — is exactly the kind of contribution this kernel wants.

We didn't take it on faith: we reproduced the branch in an isolated worktree (all 9 gates green, +6 tests), reverted each fix in isolation and confirmed its test goes **red** (so none of your tests are decoration), and ran two independent adversarial reviews over the security-critical parts. Everything here is strictly better than the current tree. **We're merging this.**

Two small completions we're taking as fast-follows on our side (tracked as `ccsc-x0t.1` and `ccsc-x0t.2` — not blockers, listed so you can see where the work leads):

1. **`thread_ts` in the sibling predicate.** It's honored in `evaluate()`/`matchApplies()` (great), but `matchSubsetOrEqual()` (`policy.ts:646`) still ignores it while `matchesIntersect()` (`policy.ts:698`) already honors it — so `detectShadowing()`, which runs at boot *and* inside the operator authoring tool `scripts/policy-validate.ts`, false-warns "rule shadowed" on two thread-disjoint rules. It's fail-**closed** (over-warns, never a bypass); the fix is a one-line mirror of `matchesIntersect`.
2. **Doc scope.** Scope the CHANGELOG's "defeats head-truncation" to signed + key-verified chains. Your code comment already notes the v1-downgrade residual; on an *unsigned* journal an attacker can renumber the survivors to `seq=1` and recompute the bare SHA-256 chain with no secret, so the pin's teeth come from the signature layer.

Plus the pre-existing follow-ups we mentioned (the `input:{}` fail-open at the production `evaluate()` gate that we're prioritizing, the `lastActiveThread` global, and the journal anchor/rotation story). Thank you again — a strong, careful first contribution, and it's in.


- Jeremy Longshore
intentsolutions.io

PR #268, opened 2026-07-13
fix(boot): start the MCP stdio transport before Socket Mode — never block the handshake on Slack

### LINKED ISSUE #266: main() blocks the MCP stdio handshake behind socket.start() — Slack Socket Mode slowness times out the 30s connect and the channel stays dead
## Summary

`main()` awaits `socket.start()` **before** `mcp.connect(transport)` (server.ts:3801-3807 on main). The MCP stdio handshake has no external dependency, but it can't begin until Slack's Socket Mode connection is up. When Socket Mode is slow — throttled reconnects after connection churn, flaky WebSocket pongs — Claude Code's 30s MCP connect window expires, the client logs a timeout, and it does **not** retry. The channel is then dead until a human reconnects it.

## Evidence (production, 2026-07-13 — two boots in a row)

```
22:40:48 Starting connection with timeout of 30000ms
22:41:18 Connection timeout triggered after 30002ms (limit: 30000ms)
22:41:18 Server stderr: [slack] bot identity: { botUserId: '...', selfBotId: '...' }
[WARN]  socket-mode:SlackWebSocket:1 A pong wasn't received from the server before the timeout of 5000ms!
22:41:18 Connection failed: MCP server "plugin:slack-channel:slack" connection timed out after 30000ms
```

Note the server printed its bot identity but never reached 'MCP server running on stdio' — it was stuck inside `socket.start()`. Running the same build standalone seconds later connected to Slack instantly, so this is a transient-Slack-slowness class of failure, exactly what a boot path shouldn't be hostage to.

Environment: macOS, Claude Code 2.1.207, spawned via the plugin's .mcp.json (tsx), Node 26.5.0.

## Fix

Connect the stdio transport first, then start Socket Mode asynchronously with bounded-backoff retries (the @slack/socket-mode client auto-reconnects once started; the retry loop only guards the initial `start()`). Outbound tools use the HTTPS WebClient and work regardless of Socket Mode state; only inbound events wait on the socket. With this ordering the MCP handshake completes in ~400ms every time. PR incoming.

Related: #265 (the other production failure we hit the same day).

CraigVG:
## What

Reorders `main()` so `mcp.connect(transport)` runs before Slack's `socket.start()`, and starts Socket Mode asynchronously with bounded-backoff retries (2s doubling to a 60s cap) guarding the initial `start()`.

## Why

Fixes #266. The stdio handshake has no external dependency, but upstream ordering held it hostage to Slack: when Socket Mode was slow (throttled reconnects after churn, flaky WebSocket pongs), Claude Code's 30s MCP connect window expired and the client gave up **without retrying** — the channel stayed dead until a human reconnected it. We hit this twice in a row in production on 2026-07-13; the server printed its bot identity, then only pong warnings, and never reached 'MCP server running on stdio'.

## How

- stdio transport connects first — with this ordering the handshake completes in ~400ms every time (six consecutive production boots observed at 404-440ms).
- `socket.start()` moves into a fire-and-forget async loop: try, on failure log to stderr and retry with backoff. The @slack/socket-mode client auto-reconnects once started, so the loop only guards the initial call.
- Window semantics: between stdio-up and socket-up, outbound tools (reply/react/…) already work — they use the HTTPS `WebClient`, not the socket. Only inbound events wait for Socket Mode, which is exactly the component that was slow anyway. Nothing new is exposed while Slack connects: gating/config all initialize before `main()`'s connect section.

Tradeoff called out: a fatal, non-transient `socket.start()` error (e.g. revoked app token) now surfaces as an endless retry loop logged to stderr instead of a crashed process. Operators see the loop in the MCP log either way; happy to fail-fast after N attempts instead if you prefer.

---

## Security impact

- [x] Changes the Socket Mode / MCP stdio transport in `server.ts`

Threat-model reasoning: no new trust boundary. The change is pure ordering — both connections still come up with identical configuration; inbound events still pass the full gate; no attacker-controlled input can influence the retry loop (its only input is Slack's connect success/failure). What an attacker-controlled message can do after this change is unchanged: messages only flow once Socket Mode is up, through the same `gate()`. Existing defense layers are untouched.

---

### Tier C — Human evidence you ran yourself

- [x] Exact command(s) you ran, copy-pasteable
- [x] Your actual terminal output pasted below
- [x] Environment: macOS, Bun 1.3.8, Node 26.5.0 (tsx spawn), Claude Code 2.1.207

Standalone boot with the fix (stdio no longer waits for Slack):

```
$ timeout 15 node_modules/.bin/tsx server.ts
[slack] bot identity: { botUserId: 'U...', selfBotId: 'B...', selfAppId: '' }
[slack] MCP server running on stdio     <-- now immediate, before Socket Mode
```

Before the fix (production Claude Code MCP log, twice in a row):

```
22:40:48 Starting connection with timeout of 30000ms
22:41:18 Connection timeout triggered after 30002ms (limit: 30000ms)
22:41:18 Server stderr: [slack] bot identity: ...
[WARN]  socket-mode:SlackWebSocket:1 A pong wasn't received from the server before the timeout of 5000ms!
22:41:18 Connection failed: MCP server "plugin:slack-channel:slack" connection timed out after 30000ms
```

After deploying: six consecutive successful handshakes at 404-440ms, running in production since 2026-07-13.

`bun test` (1251 pass) and `bun run typecheck` pass locally. The real 30s-handshake race can't be exercised from the unit suite without faking both transports, which would test the mock rather than the ordering.

---

## Checklist

- [x] I read `CLAUDE.md` and `SECURITY.md` and the change is consistent with them
- [x] Pure logic went into `lib.ts`; stateful wiring stayed in `server.ts` (boot wiring only)
- [x] I did not add runtime dependencies
- [x] Commit messages describe *why*, not just *what*
- [x] Not a breaking change
- [x] I re-read the full PR diff before submitting

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -3797,15 +3797,40 @@ async function main(): Promise<void> {
     console.error('[slack] Failed to resolve bot identity:', err)
   }
 
-  // Connect Socket Mode (Slack ↔ local WebSocket)
-  await socket.start()
-  console.error('[slack] Socket Mode connected')
-
-  // Connect MCP stdio (server ↔ Claude Code)
+  // Connect MCP stdio (server ↔ Claude Code) FIRST. The stdio handshake
+  // has no external dependency and must come up immediately: when
+  // socket.start() was awaited before mcp.connect(), any Slack-side
+  // Socket Mode slowness (throttled reconnects after connection churn,
+  // flaky WebSocket pongs) blew Claude Code's 30s MCP handshake window,
+  // and the client logged a connection timeout and gave up without
+  // retrying — the whole channel stayed dead. Outbound tools
+  // (reply/react/...) use the HTTPS WebClient and work regardless of
+  // Socket Mode state; only inbound events wait on the socket.
   const transport = new StdioServerTransport()
   transport.onclose = () => void shutdown('stdio transport closed')
   await mcp.connect(transport)
   console.error('[slack] MCP server running on stdio')
+
+  // Connect Socket Mode (Slack ↔ local WebSocket) asynchronously with
+  // bounded-backoff retries. The @slack/socket-mode client auto-reconnects
+  // once started; this loop only guards the initial start() call.
+  void (async () => {
+    let delayMs = 2_000
+    while (true) {
+      try {
+        await socket.start()
+        console.error('[slack] Socket Mode connected')
+        return
+      } catch (err) {
+        console.error(
+          `[slack] Socket Mode start failed (retrying in ${Math.round(delayMs / 1000)}s):`,
+          err instanceof Error ? err.message : err,
+        )
+        await new Promise((r) => setTimeout(r, delayMs))
+        delayMs = Math.min(delayMs * 2, 60_000)
+      }
+    }
+  })()
 
   // Belt-and-suspenders: the SDK's StdioServerTransport doesn't listen for
   // stdin end/close, so transport.onclose never fires on its own. Hook stdin
--- a/server.ts
+++ b/server.ts
@@ -43,6 +43,7 @@ import {
   EVENT_DEDUP_TTL_MS,
   enforceAuditReceiptCap,
   escMrkdwn,
+  extractSlackErrorCode,
   formatVerifyResult,
   type GateResult,
   isDuplicateEvent,
@@ -55,6 +56,7 @@ import {
   gate as libGate,
   listSessions as libListSessions,
   makeIdempotentSend,
+  NON_RETRYABLE_SLACK_ERRORS,
   PERMISSION_REPLY_RE,
   type PendingPolicyApproval,
   parseSendableRoots,
@@ -260,6 +262,14 @@ const web = new WebClient(botToken)
 const socket = new SocketModeClient({ appToken })
 
 let botUserId = ''
+// Settles once the boot-time web.auth.test() attempt completes (success OR
+// failure) — MCP connects before identity resolves, so tools that consume
+// identity (publish_manifest's replace-sweep) bounded-await this latch instead
+// of silently operating with '' identity during the window.
+let settleIdentity: () => void = () => {}
+const identitySettled = new Promise<void>((r) => {
+  settleIdentity = r
+})
 let selfBotId = ''
 let selfAppId = ''
 
@@ -1991,6 +2001,25 @@ async function executePublishManifest(
 
   // Gate 2: channel must be opted in, same as any outbound write.
   executePublishManifestGate2(channel, callerUserId, ctx)
+
+  // Identity guard: MCP connects before web.auth.test() resolves, so there is
+  // a window (sub-second happy path; up to ~30 min while Slack auth degrades
+  // and the WebClient retries) where tools are live but botUserId is still ''.
+  // findOurPriorManifestPins fails closed on '' and the replace-sweep would
+  // silently no-op, leaving duplicate pinned manifests. Bounded-await the
+  // identity latch; if identity is still unresolved, fail the call loudly as
+  // retryable rather than publish with a silent sweep skip.
+  if (ctx.botUserId === '') {
+    await Promise.race([identitySettled, new Promise((r) => setTimeout(r, 5_000))])
+    // Refresh from module state — this ctx was built before the latch settled.
+    ctx.botUserId = botUserId
+    ctx.selfBotId = selfBotId
+    if (ctx.botUserId === '') {
+      throw new Error(
+        'publish_manifest: bot identity not yet resolved (Slack auth still connecting); retry shortly',
+      )
+    }
+  }
 
   ctx.journalWrite({
     kind: 'gate.outbound.allow',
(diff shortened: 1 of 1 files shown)


CraigVG:
Pushed a hardening commit addressing three findings from an adversarial review pass on this branch:

1. **`web.auth.test()` could still stall the handshake** — the WebClient's default retry policy is ~10 retries over ~30 minutes, so awaiting identity resolution before `mcp.connect()` had the same failure mode this PR set out to fix. Identity resolution now runs inside the async Slack bring-up, before `socket.start()` — it's only consumed by inbound-event processing (mention detection, self-echo filtering), and no inbound event can arrive until the socket is up.
2. **Unrecoverable start errors now fail loud** — `not_authed` / `invalid_auth` / `account_inactive` / `user_removed_from_team` / `team_disabled` (the set the SDK throws out of `retrieveWSSURL` rather than reconnecting internally) exit non-zero via `shutdown()` instead of retrying forever. This resolves the tradeoff flagged in the PR description in favor of fail-fast.
3. **Retry loop is shutdown-aware** — `while (!shuttingDown)` so a sleeping retry can't call `socket.start()` after shutdown began and resurrect a socket in a process about to exit.

`bun test` (1251) + typecheck + coverage floor still green.

jeremylongshore:
@CraigVG — thank you for this one, and for the hardening commit you already pushed on the branch (making `not_authed` / `invalid_auth` / `account_inactive` / `user_removed_from_team` / `team_disabled` fail loud via `shutdown()` instead of retrying forever — exactly the right call, and a nice catch on running identity resolution *inside* the async bring-up so `web.auth.test()`'s own ~30-min retry can't re-block the handshake). The core insight — bring MCP stdio up first so a slow Slack side can't blow Claude Code's 30s handshake window — is correct and important; we confirmed the failure mode independently.

We reproduced the branch (all 9 gates green) and traced the boot path end to end. **We're merging this** — it's a strict improvement over `main`. Two things we're tracking as fast-follows on our side (not blockers; beads `ccsc-x0t.4` and `ccsc-x0t.3`):

**1. Round out the unrecoverable-error classification** — building on your fail-loud commit *and* Greptile's note above. `UNRECOVERABLE_START_RE` matches on `err.message` and covers only the 5 auth codes, but (a) the repo already has a canonical `NON_RETRYABLE_SLACK_ERRORS` set (`lib.ts:296`, matched via `extractSlackErrorCode` on `err.data.error`) that also names `token_revoked` / `token_expired` / `missing_scope` / `no_permission`, and (b) `SocketModeClient.retrieveWSSURL` *also* throws `RequestError` / `HTTPError` (via `err.code`) as unrecoverable-to-start — a persistent 5xx / proxy blackhole / DNS or TLS failure. Reusing `NON_RETRYABLE_SLACK_ERRORS` plus an `err.code` check for RequestError/HTTPError, instead of the message regex, makes the fail-loud set complete and consistent with the rest of the codebase.

**2. Empty-identity window for `publish_manifest`** — because MCP now connects before `web.auth.test()` resolves, there's a window (sub-second on the happy path, up to ~30 min while Slack auth is degrading and the WebClient retries) where MCP tools are live but `botUserId`/`selfBotId` are still `''`. A `publish_manifest` then → `findOurPriorManifestPins` returns `[]` (fail-closed, `manifest.ts:454`) → the replace-sweep silently no-ops → duplicate pinned manifests until the next publish. Deferring + logging the sweep when identity is unresolved closes it.

Really appreciate how fast you iterated here. Merging now; the two items above are on us.


- Jeremy Longshore
intentsolutions.io

CraigVG:
Thanks for the careful review, @jeremylongshore — both asks are landed in `f01df8a`:

**1. Bounded retry** — the loop now distinguishes three outcomes:
- auth/config-fatal: the structured `err.data.error` code is checked against `NON_RETRYABLE_SLACK_ERRORS` first (which also brings in `token_revoked` / `token_expired`, per Greptile's note), with the message regex kept as fallback for wrapped errors → immediate `shutdown(.., 1)`.
- anything else (persistent 5xx / proxy blackhole / DNS / TLS — the `RequestError`/`HTTPError` class you flagged): `MAX_SOCKET_START_ATTEMPTS = 10` (~5 minutes with the existing backoff), then `shutdown('Socket Mode start exhausted retries', 1)`. Transient-outage recovery preserved; a permanently-dead channel now always fails loud.
- `shuttingDown` still short-circuits every iteration.

**2. Empty-identity window** — went with a variant that's slightly stronger than either option: a module-level identity latch settles when the boot-time `web.auth.test()` attempt completes (success **or** failure), and `publish_manifest` bounded-awaits it (5s) when `botUserId === ''`, refreshing the call's ctx from module state after the latch. If identity is still unresolved (Slack auth degrading), the call fails loudly as retryable — so the replace-sweep can never silently no-op into duplicate pinned manifests, and the boot path still never stalls. Happy to switch to the defer-and-log approach if you'd rather the publish succeed in that window.

All gates green locally: typecheck, biome, 1251 tests, coverage floor 96.65%/98.17%, depcruise.

Issue #269, opened 2026-07-14
Kernel hardening: what the #264/#267/#268 review surfaced (and what we're fixing next)


jeremylongshore:
## The short version

Three pull requests landed close together from two people who'd never merged here before — [@parciv4l](https://github.com/parciv4l) (#264) and [@CraigVG](https://github.com/CraigVG) (#267, #268) — and all three reached straight into the parts of this project that matter most: the policy gate, the audit journal, and the boot/transport path. This issue is the story of what we did with them, what they contributed, and — the important part — the deeper backlog their PRs *illuminated* that's now ours to fix.

**First: thank you.** #264, #267, and #268 are careful, real, well-tested contributions to a security kernel from first-time contributors. That's rare and it's appreciated. **All three are merged.** This issue is the work behind our own door that they exposed.

## How it started

Because these PRs touch the exact surfaces the product exists to defend — `assertOutboundAllowed()`, the policy `evaluate()` path, `verifyJournal()`, and the MCP stdio boot — we treated both contributors as **cold** and ran a risk-mitigation review: default to *not* merging, let each PR earn merge by surviving scrutiny. CI-green means "compiles + no known-CVE + no leaked secret + a reviewer didn't object" — necessary, not sufficient, when the tests are written by the same hand as the fix.

## What we did (the review)

- **Reproduced all three independently** in isolated `git worktree`s off each PR head — not trusting the badge. All 27 gates (9 × 3) pass in a clean tree.
- **Repro'd-or-refuted every candidate risk** with runnable evidence rather than adjudicating by reading.
- **Hazard-bound #264's tests**: reverted each fix in isolation and confirmed its regression test goes *red* — so none of them are decoration.
- **Ran two independent adversarial reviewers** (a security auditor and an append-only-log specialist) over #264's security-critical parts, so the verdict isn't one person's re-read. Both converged.
- **Read the full PR threads**, including the Greptile reviews and the contributors' own hardening commits — several findings below build directly on those.

## The contributors' work — merged, with thanks 🙏

- **#267 — [@CraigVG](https://github.com/CraigVG)** (`f5b7d3e`) — route the supervisor's structured logs to **stderr** because stdout is the MCP protocol channel. Premise confirmed exactly: the supervisor runs with the default log sink, which was writing to stdout and corrupting the JSON-RPC transport on every message. Clean fix; independently necessary even alongside #268.
[...]

PR #271, opened 2026-07-14
fix(policy): close the input-unavailable fail-open — deny/require fail safe to a human at the gate (ccsc-x0t.5)

jeremylongshore:
## What

Closes the **fail-open** in the policy engine at its *sole* production call site — the crown-jewel finding of the cold-contributor review of PRs #264/#267/#268 (`ccsc-x0t.5`, under epic #269).

The MCP `permission_request` notification carries only an `input_preview` **string**, never the structured tool args, so `server.ts` built the `ToolCall` with `input: {}`. A `deny`/`require_approval` rule scoped by `pathPrefix`/`argEquals` therefore read empty input, `matchApplies()` returned **no-match**, the rule was silently skipped, and a later broad `auto_approve` swallowed the call.

**Proven exploit** (both adversarial reviewers reproduced it):

```json
[
  { "id": "deny-etc",   "effect": "deny",        "match": { "pathPrefix": "/etc" }, "reason": "no /etc" },
  { "id": "trust-bash", "effect": "auto_approve", "match": { "tool": "Bash" } }
]
```

A `Bash` call to `/etc/shadow` returned `{ allow, rule: 'trust-bash' }` — the operator's `/etc` deny was never evaluated and no human was asked. `detectBroadAutoApprove` can't catch it (the `auto_approve` is `tool`-scoped, so not "broad").

## Why + decision rationale

The acceptance criterion offered two paths: **(a)** carry structured input to the gate, or **(b)** reject/warn `pathPrefix`/`argEquals` rules at load. The MCP surface can't carry structured args today, and *silently rejecting an operator's authored guardrail is itself a fail-open*.

**Chose fail-safe-to-human over rejecting-rules-at-load** because it degrades an unconfirmable guardrail to "ask a human," never to "allow" — the kernel ethos (honest, fail-closed, small), and it reuses the existing `require_approval` primitive rather than adding a new policy effect. The linter (path b) still ships, but as a **warning**, not a silent drop.

## Layers touched + invariant changes

- `policy.ts` — `ToolCall`, `PolicyDecision` (`require` variant), `matchApplies` (now tri-state), `evaluate()`, new `detectUnenforceablePredicates()` linter, three fail-safe constants.
- `lib.ts` — `PolicyDecisionShape.require` mirrors the new optional `reason` (the bidirectional `satisfies` shape-drift guard stays green).
- `server.ts` — production gate sets `inputAvailable: false`; boot linter wired into `loadPolicyRulesAtBoot`; honest `failsafeReason` merged into the `policy.require` journal echo.
- **New invariants:** `policy-evaluation-flow.md` #9 and `THREAT-MODEL.md` #8 — *the policy gate never fails open*; an unevaluable predicate degrades to "ask a human," never "allow."

## How it works

1. `ToolCall` gains `inputAvailable?: boolean` (default **true** → every existing call site and test byte-for-byte unchanged). The production gate sets it `false`.
2. `matchApplies()` is **tri-state** (`match` | `no_match` | `indeterminate`). Enforceable fields (tool/channel/thread_ts/actor) are checked **first** and short-circuit to `no_match`; only when they all match and an input-dependent predicate is unevaluable does it return `indeterminate`.
3. `evaluate()` on `indeterminate`: **deny/require_approval → `require`** (routes to a human, preempting any later `auto_approve`); **auto_approve → skip** (never auto-approve on an unconfirmable predicate). A fresh `(rule, session)` approval still flips it to `allow`.
4. **Honest journaling:** the fail-safe `require` carries `reason: 'predicate unevaluable at gate → routed to human'`, stamped into `policy.require` as `failsafeReason`. No `policy.allow` is ever written — the signed chain never claims the predicate was evaluated.
5. **Boot linter** `detectUnenforceablePredicates()` (warn-not-block) names every `pathPrefix`/`argEquals` rule and states its effective behavior per effect.

## Verification & evidence

- **15 regression tests** added (`server.test.ts`): the `/etc/shadow` exploit → `require` not `allow`; argEquals fail-safe; indeterminate `require_approval` uses the rule's own ttl/quorum; indeterminate `auto_approve` skipped + a later deny still wins; a deny scoped to a *different* tool does **not** spuriously fail-safe; a fresh approval flips indeterminate→allow; `inputAvailable` defaults true → prior behavior byte-for-byte; `failsafeReason` reaches the `policy.require` event; full linter coverage.
- **Revert→red proven:** restoring the fail-open makes the crown-jewel test return `{allow, rule:'trust-bash'}` instead of `{require, rule:'deny-etc'}`.
- **All nine CI gates green locally:** typecheck · biome · `bun test` (1272 pass / 0 fail) · coverage 96.66% line / 98.21% func (floor 95) · depcruise · gherkin-lint · harness-hash · bun audit · crap-score (`policy.ts` max 23 < 30). CI on this PR is the authoritative run.

## Risk assessment

The fail-safe is **coarser** than a true evaluation — a `/etc` deny becomes "ask a human about this `Bash` call" rather than "block paths under `/etc`." A human could approve a call the precise predicate would have blocked. This is an honest limitation of the current MCP surface (no structured args at the gate), journaled as such, and **strictly safer than the silent allow it replaces**. It resolves fully when the `permission_request` notification carries structured input (`inputAvailable` becomes `true` at that site; predicates evaluate precisely; the fail-safe path is never taken).

## Operational impact

None required. Additive and backward-compatible: no config/env/migration/dep changes. Operators with `pathPrefix`/`argEquals` rules will see new **warn-not-block** boot lines from `detectUnenforceablePredicates` describing the effective gate behavior; nothing fails to boot.

## Follow-up & deferred

Sibling epic children remain, each shipping as its own small PR: `ccsc-x0t.1`/`.2`/`.3`/`.4` (contributor fast-follows), `.6` (`lastActiveThread` cross-thread race), `.7`/`.9` (audit tamper-evidence), `.8` (`getChannelPolicy` helper), `.10` (boot/transport integration tests). Tracked under epic #269.

## Governance links

- Design: [`000-docs/policy-evaluation-flow.md`](000-docs/policy-evaluation-flow.md) § Input-unavailable fail-safe + invariant 9
- Threat model: [`000-docs/THREAT-MODEL.md`](000-docs/THREAT-MODEL.md) T12 + invariant 8

Refs #269


jeremylongshore:
Merged (squash 49915d4). Settles the crown-jewel item **ccsc-x0t.5** of epic #269 — the input-unavailable policy fail-open is closed.

- Nine-gate CI green (Typecheck 1272 pass/0 fail, coverage 96.66% line / 98.19% func), CodeQL + gitleaks pass, Gemini review clean.
- Revert→red proven on the `/etc/shadow` crown-jewel test.
- Design-in-public: `policy-evaluation-flow.md` §Input-unavailable fail-safe + invariant 9; `THREAT-MODEL.md` T12 + invariant 8; CHANGELOG.

Continuing epic #269 (kernel hardening) with the remaining children as follow-on PRs.

- Jeremy Longshore
intentsolutions.io

PR #267, opened 2026-07-13
fix(supervisor): route structured logs to stderr — stdout is the MCP protocol channel

CraigVG:
## What

Routes the supervisor's default structured-log sink from stdout to stderr. One word changed (`process.stdout.write` → `process.stderr.write`), plus the doc comment updated to say why stdout is off-limits.

## Why

Fixes #265. `server.ts` never passes a custom `log` to `createSessionSupervisor`, so `defaultLog` is the production sink — and the server speaks MCP over stdio, so stdout is the protocol channel. Every `session.activate` line (one per inbound message) and every boot-time recovery-sweep line arrives at Claude Code as an invalid JSON-RPC message and the client drops the whole connection: reconnect churn on every message, lost in-flight notifications, and 30s handshake timeouts at boot when the sweep has files to classify.

## How

stderr is where every other diagnostic in this codebase goes (`console.error` throughout `server.ts`), and MCP hosts capture child stderr harmlessly, so the structured-log format is preserved for anything tailing it. Callers who want a different sink still pass their own `log` — the injection point is unchanged.

---

## Security impact

- [x] None of the above — this is a docs / test-only / cosmetic change

(One logging sink redirected; no gating, transport, tool, or token logic touched. Arguably it *hardens* the transport: supervisor state objects no longer traverse the protocol channel.)

---

### Tier C — Human evidence you ran yourself

- [x] Exact command(s) you ran, copy-pasteable
- [x] Your actual terminal output pasted below
- [x] Environment: macOS, Bun 1.3.8, Node 26.5.0 (tsx spawn), Claude Code 2.1.207

Before (production Claude Code MCP log, every inbound Slack message):

```
22:27:16.034 STDIO connection dropped after 285s uptime
22:27:16.034 Connection error: [ { "code": "invalid_union", ... path: ["jsonrpc"] ...
  { "code": "unrecognized_keys", "keys": ["event", "channel", "thread", "ownerId"] } ...
```

After (same session, fix deployed): connection stable across inbound messages, no drops in the MCP log since 2026-07-13 22:44 — running in production since.

`bun test` (1251 pass) and `bun run typecheck` pass locally. No automated test added: `defaultLog` is module-internal by design (the injection point is the `log` option), and asserting on a real process's stdout/stderr split from inside the suite would test the harness more than the code.

---

## Checklist

- [x] I read `CLAUDE.md` and `SECURITY.md` and the change is consistent with them
[...]

PR #274, opened 2026-07-14
refactor(boot): extract socket-start + manifest-identity guards to testable lib.ts seams and bind #267/#268 (ccsc-x0t.10, closes ccsc-x0t.3/.4)

### LINKED ISSUE #269: Kernel hardening: what the #264/#267/#268 review surfaced (and what we're fixing next)

jeremylongshore:
## What

#267 (structured logs → stderr) and #268 (MCP-before-Slack ordering + Socket-Mode bounded retry + empty-identity `publish_manifest` guard) each shipped with **zero** new tests, and their logic lived in `server.ts`'s `main()`-scope boot wiring — which runs on import, so it can't be unit-tested directly. A later refactor could silently revert either fix. This binds them, and closes `ccsc-x0t.10` (and the code-ACs of `ccsc-x0t.3` + `ccsc-x0t.4`) under epic #269.

## Why + decision rationale

The boot wiring isn't importable (server.ts runs `main()` + constructs Slack clients on import — nothing in the suite imports it). **Chose to extract the pure *decisions* into `lib.ts` over importing server.ts (impossible) or leaving the fixes untested** — the untested gap is exactly what let two cold contributors find these bugs. The extraction is behavior-preserving; `server.ts` calls the helpers.

## Layers touched

- `lib.ts` (pure kernel) — **additive** exports; no change to any existing export. Not in AGP's vendored relay subset (socket-start + manifest-identity are CCSC-Slack-boot-specific).
- `supervisor.ts` — `export` on `defaultLog` (no logic change).
- `server.ts` (boot path — security-sensitive) — behavior-preserving refactor: the Socket-Mode retry loop and the `publish_manifest` identity guard now call the extracted helpers.

## How it works

- `classifySocketStartError(err)` → `'fatal'` (auth/config — reuses `NON_RETRYABLE_SLACK_ERRORS` + a `UNRECOVERABLE_SOCKET_START_RE` message fallback) or `'retryable'` (transient). Drives the loop's shut-down-vs-retry decision.
- `nextSocketStartBackoffMs(prev)` → exponential, capped at `SOCKET_START_BACKOFF_CAP_MS` (60s).
- `assertManifestIdentityResolved(botUserId)` → throws `MANIFEST_IDENTITY_UNRESOLVED_MSG` on `''`. Publishing before Slack auth resolves would make `findOurPriorManifestPins` fail closed and the replace-sweep silently no-op, leaving duplicate pinned manifests.
- `defaultLog` exported so a test asserts it writes to **stderr, never stdout** (a structured line on stdout is an invalid JSON-RPC frame that kills the MCP transport — the #267 regression).

## Verification & evidence

- **7 regression tests** (`server.test.ts`): `defaultLog` → stderr never stdout; `classifySocketStartError` fatal-on-code / fatal-on-message-regex / retryable-on-transient; `nextSocketStartBackoffMs` doubles-then-caps; `assertManifestIdentityResolved` empty-throws / resolved-ok.
- **All nine gates green locally:** typecheck · biome · `bun test` (1285 pass / 0 fail) · coverage 96.66% line / 98.21% func · depcruise · gherkin-lint · harness-hash · bun audit · crap-score (`server.ts` max 29 < 30). CI on this PR is authoritative.

## Risk assessment

Behavior-preserving refactor of a security-sensitive boot path. The classification + backoff are byte-identical to #268's inline logic (now shared, not duplicated); the identity-guard throw is unchanged (same message, exported as a constant). The loop/shutdown wiring stays in `server.ts` (typecheck-covered, consistent with its other `main()`-scope code — not unit-tested because the module isn't importable). `lib.ts` additions are additive. Backward-compatible; no config/API/env change.

## Operational impact

None. No env/migration/dep changes.

## Follow-up & deferred

- `ccsc-x0t.3` (empty-identity guard) and `ccsc-x0t.4` (bounded-retry-then-`shutdown(1)`) — code shipped in #268; **closed by the binding tests here**.
- Remaining epic children under #269: `.6` (thread correlation — design decision, blocked on notification shape), `.7`/`.9` (journal-verify hardening), `.8` (`getChannelPolicy` helper).

## Governance links

- Design: [`000-docs/session-state-machine.md`](000-docs/session-state-machine.md) (supervisor), [`000-docs/bot-manifest-protocol.md`](000-docs/bot-manifest-protocol.md) (publish guard)

Closes #269 partially — Refs #269


jeremylongshore:
Thanks — all three findings addressed in 140d3a8 (they're genuine hardening on security-adjacent boot code):

1. **`classifySocketStartError` message extraction** — added a private `errorMessage(err)` helper that reads an `Error`'s `.message` *or* a plain error-like object's string `.message` (else `String(err)`), so a wrapped/serialized fatal error thrown as a plain object no longer stringifies to `"[object Object]"` and slip past the regex → retryable. Bound by a new `{ message: 'token_revoked' }` → `fatal` test.
2. **`assertManifestIdentityResolved`** — signature widened to `string | null | undefined` and the check is now `!botUserId || botUserId.trim() === ''`, so the guard fails closed on every invalid identity, not just the exact `''`.
3. **Tests expanded** — the guard now asserts throw on `''`, `null`, `undefined`, and `'   '`.

All nine gates green locally (test 1285 pass / 0 fail).

- Jeremy Longshore
intentsolutions.io

PR #275, opened 2026-07-14
refactor(access): route all channel-policy lookups through one fail-closed getChannelPolicy helper (ccsc-x0t.8)

jeremylongshore:
## What

Channel-policy reads (`access.channels[id]`) were scattered across **seven sites** with inconsistent guards — some hardened with `Object.hasOwn` (#264), others a bare `access.channels[id]` index. A bare index reads inherited `Object.prototype` members for ids like `'constructor'`/`'toString'`/`'__proto__'` (a truthy *function* off the prototype chain), which could slip past a truthiness-based gate. `ccsc-x0t.8` from the #264 review (epic #269).

## Why + decision rationale

Not exploitable today (Slack channel ids are `[CD][A-Z0-9]+`, never prototype-key forms), but a single fail-closed accessor removes the footgun class **and** the six scattered ad-hoc guards. **Chose one shared helper over hardening each site in-place** because the value is having exactly one chokepoint that can't drift as new call sites are added.

## Layers touched

- `lib.ts` — new `getChannelPolicy(access, id)`; three call sites rewired (`assertOutboundAllowed`, peer-bot gate, inbound channel gate).
- `server.ts` — imports the helper; four call sites rewired (audit-receipt, `perUserSessions`, admin-command authz, and the one already using `?.[`).

## How it works

```ts
export function getChannelPolicy(access: Access, channelId: string): ChannelPolicy | undefined {
  return Object.hasOwn(access.channels, channelId) ? access.channels[channelId] : undefined
}
```

Returns the policy only for an **own-property** id, else `undefined`. Every gate/read now goes through it.

## Verification & evidence

- **3 tests** (`server.test.ts`): own-id returns the policy; missing-id `undefined`; a prototype-key sweep (`constructor`/`toString`/`hasOwnProperty`/`__proto__`/`valueOf` all → `undefined`).
- **Revert→red proven:** a bare `access.channels[id]` reddens the prototype-key test.
- **All nine gates green locally:** typecheck · biome · `bun test` (1288 pass / 0 fail) · coverage 96.66% line / 98.21% func · depcruise · gherkin-lint · harness-hash · bun audit · crap-score. CI on this PR is authoritative.

## Risk assessment

Behavior-preserving for all real channel ids (`Object.hasOwn` is true for any authored channel). The only behavior change is that prototype-key ids now return `undefined` instead of a prototype member — strictly fail-closed. `lib.ts` sites call a same-module hoisted function; `server.ts` imports it. Additive; no config/API/env change.

## Operational impact

None. No env/migration/dep changes.

## Follow-up & deferred

Remaining epic #269 children: `ccsc-x0t.6` (thread correlation — design decision, blocked on the MCP notification shape), `ccsc-x0t.7` (journal v2-floor anchor + rotation doc), `ccsc-x0t.9` (`eventsVerified` floor so a wiped log ≠ "verified clean"). Plus a small follow-up applying two Gemini test-hygiene suggestions on the #273 forgery tests.

## Governance links

- Design: [`ACCESS.md`](ACCESS.md) (access schema), [`000-docs/THREAT-MODEL.md`](000-docs/THREAT-MODEL.md) (T5 state-file / prototype-chain hardening)

Refs #269


jeremylongshore:
Good catch — fixed in 5a2b3c4. `getChannelPolicy` now guards `access.channels` before `Object.hasOwn`:

```ts
const channels = access.channels
return channels && Object.hasOwn(channels, channelId) ? channels[channelId] : undefined
```

A loaded-from-disk `Access` can lack `channels` (which is exactly why the admin-command site used `channels?.[id]` before this refactor), and `Object.hasOwn(undefined, …)` throws — so the missing-channels case now fails closed (returns `undefined`) instead of crashing. Added a regression test asserting it doesn't throw and returns `undefined`; revert→red verified. All nine gates green (1289 pass / 0 fail).

- Jeremy Longshore
intentsolutions.io

PR #276, opened 2026-07-14
test(journal): apply Gemini test-hygiene suggestions from the #273 review (ccsc-xkf)

jeremylongshore:
## What

Two Gemini inline comments on **#273** posted *after* that PR merged (surfaced by an all-PR Gemini audit). They're test-quality suggestions on the v1/v2 journal forgery tests: assert the setup produced 4 events before slicing (removing a `survivors[0]!` non-null assertion), and build a fresh object in `.map()` instead of mutating the destructured `rest`.

## Why

Test-quality only — the tests bind correctly and every assertion is unchanged. Honoring a reviewer comment even post-merge; makes setup failures loud and removes the unsafe `!` + the map-mutation smell.

## How it works

- `expect(events.length).toBe(4)` before `events.slice(1)` → `survivors[0].prevHash` no longer needs `!`.
- `const updated = { ...rest, seq: i + 1, prevHash }` replaces the mutation of `rest`. `canonicalJson` sorts keys, so the hashed bytes are byte-identical — the forgery still reproduces exactly.

## Verification & evidence

Both `ccsc-x0t.2` forgery tests still pass (v1 → `ok:true`, v2 → `ok:false`). All nine gates green locally: typecheck · biome · `bun test` (1289 pass / 0 fail) · coverage 96.66% line / 98.21% func · depcruise · gherkin-lint · harness-hash · bun audit · crap-score.

## Risk assessment

None — test-only refactor; hashed bytes and all assertions unchanged. No runtime code touched.

## Operational impact

None.

## Follow-up & deferred

Closes the #273 review follow-up (`ccsc-xkf`). Remaining epic #269 children: `ccsc-x0t.9` (verify `--min-events` floor), `ccsc-x0t.7` (v2-floor anchor), `ccsc-x0t.6` (thread-correlation design decision).

## Governance links

Refs #273 · Refs #269


jeremylongshore:
Respectfully declining this one — the suggestion would break CI:

- The premise ("`survivors[0].prevHash` is a type error") doesn't hold for this repo's tsconfig: `noPropertyAccessFromIndexSignature` is off, so dot access on `Record<string, unknown>` is allowed (returns `unknown`). The Typecheck gate is green on this PR, which proves it.
- More decisively, the repo's Biome gate enforces `complexity/useLiteralKeys`, which **requires** dot notation and would flag `survivors[0]['prevHash']` as an error. Applying the suggestion would fail the deterministic CI lint gate.

Keeping the dot form. Thanks for the review.

- Jeremy Longshore
intentsolutions.io

PR #277, opened 2026-07-14
feat(verify): add a --min-events floor so a wiped audit.log fails instead of reading as verified-clean (ccsc-x0t.9)

jeremylongshore:
## What

`verifyJournal()` of a wiped-to-empty `audit.log` returns `{ ok: true, eventsVerified: 0 }`, which `formatVerifyResult` printed as `OK: 0 event(s) verified` with **exit 0** — so a monitoring cron reads a *destroyed* log as "verified clean." `ccsc-x0t.9` from the #264 review (epic #269).

Reproduced: `verifyJournal('')` → `{"ok":true,"eventsVerified":0}` → `formatVerifyResult` → `OK: 0 event(s) verified … exitCode 0`.

## Why + decision rationale

Hash-chain integrity can't detect a log truncated to nothing (an empty file is a valid empty chain). **Chose an operator-supplied floor over a hard "empty = fail"** because a first-boot log legitimately has zero events, so a blanket rule would break the common case. This is a *lower-bound liveness* check layered on the integrity check.

## Layers touched

- `lib.ts` — new `parseMinEventsArg`; `formatVerifyResult` gains an optional `minEvents` param.
- `server.ts` — the `--verify-audit-log` CLI parses `--min-events` and passes it through.
- `000-docs/audit-journal-architecture.md` — documents the flag + rationale.

## How it works

`--min-events N` (space or `=` form; rejects negatives/non-integers/empty/flag-shaped). A hash-clean result whose `eventsVerified < N` returns `{ exitCode: 1, "FAIL: audit journal too short …" }`. Absent floor → byte-identical prior behavior. A genuine chain break still reports the break (the break path is checked before the floor).

## Verification & evidence

- **7 tests**: below-floor → fail; at/above → ok; omitted-floor preserves prior behavior (`0 events → OK exit 0`); a genuine break still wins over the floor message; parser space/equals/absent/malformed.
- **Revert→red proven:** removing the floor check reddens the below-floor test.
- **All nine gates green locally:** typecheck · biome · `bun test` (1296 pass / 0 fail) · coverage 96.65% line / 98.21% func · depcruise · gherkin-lint · harness-hash · bun audit · crap-score. CI on this PR is authoritative.

## Risk assessment

Backward-compatible — no floor supplied → byte-identical prior output. The floor is a **lower-bound liveness** check, NOT tamper-evidence: it detects a *shrunk* log, not a full-file forge (that's the signed-chain / v2-floor story, `ccsc-x0t.7`). Documented as exactly that, so the signed log never overclaims.

## Operational impact

None required. New optional CLI flag; existing `--verify-audit-log` invocations unchanged. Operators can add `--min-events N` to their verification cron for shrink-detection.

## Follow-up & deferred

`ccsc-x0t.7` (v2-floor / pinned-genesis anchor + rotation doc) and `ccsc-x0t.6` (thread-correlation design decision) remain under epic #269.

## Governance links

- Design: [`000-docs/audit-journal-architecture.md`](000-docs/audit-journal-architecture.md) § CLI

Refs #269


jeremylongshore:
Great catch — this is a real fail-open in a security check, fixed in 1c4261b.

`parseMinEventsArg` now **fails closed**: a present-but-malformed value throws (`invalid --min-events value …`) and a missing/flag-shaped value throws (`missing value for --min-events`), instead of returning null and silently disabling the floor. Absent flag still returns null (the floor is optional). The `--min-events` parse also moved inside the verify `try`, so a malformed flag exits non-zero with a clear message rather than crashing at module load. Tests updated to assert the throws; revert→red on the floor itself still holds. All nine gates green (1296 pass / 0 fail).

- Jeremy Longshore
intentsolutions.io

PR #278, opened 2026-07-14
feat(journal): add pinned-genesis + v2-floor verification anchors to close the head-truncation/downgrade residuals (ccsc-x0t.7)

jeremylongshore:
## What

The genesis-seq pin (#264) catches a *naive* head-shear but not the two attacks documented as the `ccsc-x0t.2` residual: **(a)** a v1 renumber+rechain (recomputable with no secret → verifies clean), and **(b)** a uniform downgrade of a signed chain to unsigned v1 (the in-file v2→v1 rollback check never fires on a uniformly-v1 file). Two optional `VerifyOptions` anchors close both. `ccsc-x0t.7`, epic #269 — the full fix for the residual PR #273 documented.

## Why + decision rationale

The only way to catch these is an **out-of-band fact the attacker can't reproduce from the file itself**. **Chose operator-supplied anchors over a self-describing scheme** because a self-attested anchor is forgeable alongside everything else — the point is that the anchor lives *outside* the file. Additive + opt-in: absent → verification is byte-identical.

## Layers touched

- `journal.ts` — `VerifyOptions` gains `pinnedGenesisHash` + `v2FloorSeq`; enforced via a new `checkVerifyAnchors()` helper (factored out to keep `verifyJournal` under the CRAP gate, mirroring `checkSeqContinuity`). **Security-contract change** → design-doc-first.
- `lib.ts` — `parseV2FloorSeqArg` + `parseExpectedGenesisArg` (fail-closed); `parseMinEventsArg` refactored to share a new `parseNonNegIntFlag` helper (behavior-identical).
- `server.ts` — `--verify-audit-log` wires `--v2-floor-seq` / `--expected-genesis-hash`.
- `000-docs/audit-journal-architecture.md` — § Verification anchors + § Rotation and slice reseed.

## How it works

| Anchor | Rule | Defeats |
|---|---|---|
| `pinnedGenesisHash` (`--expected-genesis-hash HEX`) | first accepted event's `prevHash` MUST equal it | head-shear + rechain (v1 **and** v2) |
| `v2FloorSeq` (`--v2-floor-seq N`) | every event at `seq ≥ N` MUST be v2 | uniform downgrade-to-v1 of a chain signed from N |

The genesis hash is the writer's `initialPrevHash` captured out-of-band; `v2FloorSeq` is the seq signing began at. Both parsers **fail closed** — a malformed value throws rather than silently disabling the anchor (the lesson from #277). **Rotation/reseed** is a documented operator procedure (archive old file, open new with `initialPrevHash` = old head hash; verify a slice with `--expected-genesis-hash`) — no writer change.

## Verification & evidence

- **8 tests** including the **money test**: `pinnedGenesisHash` makes the *exact* `ccsc-x0t.2` v1 forgery **FAIL** where it verified clean unanchored; `v2FloorSeq` breaks a v1 event at/after the floor; a fully-signed chain passes its floor; both parsers fail-closed.
- **Revert→red proven:** neutering the genesis check reddens the forgery-defeat test.
- **All nine gates green locally:** typecheck · biome · `bun test` (1304 pass / 0 fail) · coverage 96.65% line / 98.21% func · depcruise · gherkin-lint · harness-hash · bun audit · crap-score (`verifyJournal` back to 29 < 30 after the extraction). CI on this PR is authoritative.

## Risk assessment

Additive + opt-in — no anchor supplied → byte-identical prior verification. The anchors are only as strong as the operator's out-of-band capture of the genesis hash / signing-floor seq; documented as exactly that (a rotation that forgets to record the head hash loses the cross-file link). No config/env/dep change; new optional CLI flags. Contract-level convergent with AGP — no shared code, both runnable standalone.

## Operational impact

None required. Operators who captured the genesis hash / signing-floor seq can add `--expected-genesis-hash` / `--v2-floor-seq` to their verification cron for stronger tamper-evidence. Rotation gains a documented reseed procedure.

## Follow-up & deferred

`ccsc-x0t.6` (thread-correlation — upstream-blocked design decision) is the **last** open child of epic #269.

## Governance links

- Design: [`000-docs/audit-journal-architecture.md`](000-docs/audit-journal-architecture.md) § Verification anchors + § Rotation and slice reseed; [`000-docs/THREAT-MODEL.md`](000-docs/THREAT-MODEL.md) T8
- Changelog: `[Unreleased]` → Added

Refs #269


jeremylongshore:
Applied in fa2e716 — good consistency catch. A token that starts with `-` followed by a digit (a negative number) now routes to the value validator, so `--v2-floor-seq -5` throws the accurate `invalid … value: must be a non-negative integer` instead of a misleading `missing value`. A genuine next-flag like `--verify-audit-log` still reads as `missing value`. Applied to both the shared int-flag parser and the genesis-hash parser; test updated. Both forms still fail closed. All nine gates green (1304 pass / 0 fail).

- Jeremy Longshore
intentsolutions.io

PR #282, opened 2026-07-17
fix(security): enforce policy thread_ts + pin journal genesis seq (Fable-5 review)

parciv4l:
Lands fixes from a **Fable 5 adversarial security review** of the load-bearing core. No incident — proactive review that surfaced real latent defects before they could be hit.

## Findings fixed

- **HIGH — `policy.ts`:** `matchApplies()` now compares `match.thread_ts` against `call.sessionKey.thread`. It previously ignored the field despite the schema accepting a `thread_ts`-only rule, so a thread-scoped `auto_approve` silently applied to **every** thread and channel. Never bit (no such rule authored) but was a real permission-bypass. Doc updated: the field was reserved schema-only "until Epic 29-B wires `evaluate()`"; this is that wiring.
- **MEDIUM — `journal.ts`:** `verifyJournal()` now requires the first accepted event to carry `seq === 1`, defeating head-truncation (shearing lines off the top of the log, which otherwise verified cleanly because the first event's `prevHash` is trusted as an unpinnable per-file genesis anchor). Sequence checks extracted into `checkSeqContinuity()` to hold `verifyJournal` under the Wall-5 CRAP gate. Full-file v1 downgrade remains residual.
- **LOW — `lib.ts`:** `assertOutboundAllowed()` and the inbound channel handler use `Object.hasOwn()` for `access.channels` lookups, closing a prototype-chain truthiness read on `constructor`/`toString` channel ids. Not exploitable (Slack ids never take those forms) — defence-in-depth.
- **LOW (doc) — `THREAT-MODEL.md` §T7:** corrected `PERMISSION_REPLY_RE` keyspace (25 letters, excludes only `l`, ~9.8M). The regex is a fail-closed validator, not a generator.

## Verification

- `bun run typecheck` clean
- `bun test` — **1257 pass / 0 fail** (+4 new tests for the fixes)
- All nine CI gates green


parciv4l:
Withdrawing — these fixes are already on main (#264 plus #271–#278). Opened in error against the wrong target; apologies for the noise.

PR #283, opened 2026-07-17
feat(reliability): reconnect/startup catch-up for messages missed while disconnected

parciv4l:
Slack Socket Mode does **not** redeliver events sent while the app was disconnected, so any message the operator sent during a host sleep/reboot/crash — or a brief mid-session socket drop — was silently lost: the bridge never pulled `conversations.history` on (re)connect.

## What this adds

A per-channel **high-water mark** (`STATE_DIR/watermarks.json`, atomic `0o600` write mirroring `saveAccess`) recording the `ts` of the last message `handleMessage()` finished processing. On startup after `socket.start()` and on every subsequent `'connected'` reconnect, a one-time catch-up pulls `conversations.history` since the mark for each watermarked channel and replays missed messages through the **exact same** `handleMessage()` → `gate()` path the live socket uses.

## Security

Replay never bypasses the inbound gate, self-echo triple-check, allowlist, peer-bot filter, or admin-verb handling — a recovered message is treated identically to a live one (dropped if the gate says so). Dedup is layered: `selectCatchupMessages` drops anything at-or-before the watermark + duplicate `ts` within a batch; `handleMessage`'s `isDuplicateEvent` guard drops anything already processed live, so a message arriving live during the catch-up window is never double-handled. Per-channel history errors are swallowed and logged — catch-up is best-effort and never crashes startup.

## Structure

Pure core in `lib.ts` (`parseWatermarks` / `advanceWatermark` [monotonic] / `selectCatchupMessages` [strict > watermark, dedup, oldest→newest] / `compareSlackTs` [integer-halves compare, precision-exact]) with 27 new `bun:test` unit tests. Wiring stays in `server.ts`. Initial-connect double-run guarded (the `'connected'` listener skips the first event; a `catchupInFlight` flag serializes overlapping runs).

## Verification

- `bun run typecheck` clean
- `bun test` — **1331 pass / 0 fail**

> Note: complements the operator's separate goal of a reboot autostart — this ensures that once the bridge *is* back up, messages sent while it was down get replayed rather than lost.

### DIFF
--- a/lib.ts
+++ b/lib.ts
@@ -2171,6 +2171,124 @@ export function isDuplicateEvent(
 }
 
 // ---------------------------------------------------------------------------
+// Reconnect / startup catch-up watermarks (ccsc-reconnect-catchup)
+// ---------------------------------------------------------------------------
+//
+// Slack Socket Mode does NOT redeliver events that were sent while the app
+// was disconnected (host asleep / rebooted / crashed). The bridge therefore
+// persists a per-channel high-water mark — the `ts` of the last message it
+// finished processing in each channel — to STATE_DIR/watermarks.json. On
+// (re)connect it pulls conversations.history for everything newer than the
+// mark and replays it through the SAME gate()/handleMessage path, so nothing
+// the operator sent during the gap is silently lost. These helpers are the
+// pure, side-effect-free core (parse / advance / select); the disk I/O and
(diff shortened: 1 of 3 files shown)


parciv4l:
Withdrawing — opened without confirming this was intended as an upstream contribution. Apologies for the noise.

